PR #1, opened 2019-04-13
Add basic proof editor component

Khazuar:


src/components/proofEditor/proofEditor.module.scss
```diff
@@ -0,0 +1,15 @@
+.proof-editor {
```
Nifdee: You don't need an empty css class; and if you do need this for some reason, it should be called `proofEditor`.

src/components/proofEditor/proofEditor.module.scss
```diff
@@ -0,0 +1,15 @@
+.proof-editor {
+	
+}
+
+.textInput {
+	display: block;
+	width: 90%;
+	margin-top: 10;
```
Nifdee: What unit is this? Can you use `em` or `rem` instead?

Nifdee: Since you're setting three margin components here, you could also do `margin: 10 auto 0 auto` instead.

src/components/proofEditor/proofEditor.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+interface State {
+    text: string;
+}
+
+export class ProofEditor extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+        this.state = {
```
Nifdee: You don't need a constructor for this; instead, you can write the following:
```
public state: State = { ... };
```

src/components/proofEditor/proofEditor.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <textarea
+                className={styles.textInput}
+                value={this.state.text}
+                onChange={ev => this.setState({ text: ev.target.value })}
+            />
+            <div className={styles.buttons}>
+                <button onClick={this.checkInput}>
+                    Prüfen
+                </button>
+                <button onClick={() => { alert("Toller HInweis"); } }>
```
Nifdee: Typo in `Hinweis`

PR #9, opened 2019-04-24
Add a vocabulary checker with basic functionality and dummy dictionary

Korosensei42:
The dictionary (AllowedVocab.json) just contains some words for testing purposes.

Things to work on (maybe in a new issue (?)):

- [ ] Edit the dictionary

- [ ] Maybe change the tests for more coverage.


src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    let issue : string[] = []; //collect all wrong words in a stringarray
+    for (const word of preprocessedText) {
+        if (!json.foo.includes(word)) {//word not included in AllowedVocab.json
+            console.log(word + " ist kein erlaubtes Wort. Bitte verwenden Sie ein anderes!"); //for debugging purposes. The user should get a message on screen telling him what's wrong.
+            issue.push(word);
+        }
+    }
+    let allWrongWords :string = "";
+    allWrongWords += issue.join //Join all wrong words to one String
+    allWrongWords += " sind nicht erlaubte Wörter. Bitte verwenden Sie andere!"
```
Khazuar: Please add semicolons at the end of your statements. Technically, these are not necessary, but you can easily make mistakes this way.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log(allWrongWords); //Tell the User what's wrong
+    return issue;
+}
+
+/**
+ * Delete any unwanted punctuation
+ */
+
```
Khazuar: No new line here. Or between any documentation and the thing it documents.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+export function eliminatePunctuation(text: string){
+    var punctuationRegularExpression = /[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]/g;//delete some unicoe characters as well as specific punctuation signs.
```
Khazuar: please put that regex into a constant:
```typescript
const punctuationRegularExpression = /[\u2000-\u206F\u2E00-\u2E7F\\'!"#$%&()*+,\-.\/:;<=>?@\[\]^_`{|}~]/g;
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    var spaceRegularExpression = /\s+/g;
```
Khazuar: Same.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    text = text.replace(punctuationRegularExpression, ' ').replace(spaceRegularExpression, ' ');
+    return text;
```
Khazuar:
```typescript
return text.replace ...
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Collects all positions where there is a whitespace in a string.
+ */
+
+function findSpacePositionsInStringArray(text: string){
+    var index = 0;
+    var spacePositions = [];
```
Khazuar: no `var` please. Use `const` instead (or `let`, but only if you need to reassign the variable).

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return spacePositions;
+}
+
+/**
+ * Seperates a given string along it's whitespaces.
+ */
+
+export function transformTextToStringArray(text: string){
+    text.trim; //delete leading and trailing whitespaces
```
Khazuar: `text = text.trim();`

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    var spacePositions = findSpacePositionsInStringArray(text);
+    var tempPos = 0;
+    let result :  string[] = [];
+    spacePositions.forEach(element => {//this loop is problematic. Adds a whitespace to the beginning of each word after the second one.
+        result.push(text.substring(tempPos, element));
+        tempPos = element;
+    });
+    //the substring after the last whitespace seems to miss in general, if this substring is not already empty. The next line fixes this.
+    if(text.substring(spacePositions[spacePositions.length - 1], text.length) != " "){
+    result.push(text.substring(spacePositions[spacePositions.length - 1], text.length));
```
Khazuar: Indentation.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    for (let index = 0; index < result.length; index++) {
+        if(result[index].startsWith(' ')){
+            result[index] = result[index].substring(1, result[index].length); //delete the wrong whitespaces in the beginning
+        }
+    }
+    return result;
+}
```
Khazuar: I am not quite happy with this algorithm yet. There ought to be a simpler way of achieving this without so many loops and indexes.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    var cleantext = eliminatePunctuation(text);
+    var textarray = transformTextToStringArray(cleantext);
+    return textarray;
+}
+
+export function checkInput(text: string){
+    let preprocessedText : string[] = preprocessingText(text);
+    let issue : string[] = []; //collect all wrong words in a stringarray
+    for (const word of preprocessedText) {
+        if (!json.foo.includes(word)) {//word not included in AllowedVocab.json
```
Nifdee: `json.foo` is not that descriptive; I'd rather create a `const allowedWords = json.foo;` at the top of the file and then use that here.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function preprocessingText(text: string){
+    var cleantext = eliminatePunctuation(text);
+    var textarray = transformTextToStringArray(cleantext);
+    return textarray;
+}
+
+export function checkInput(text: string){
+    let preprocessedText : string[] = preprocessingText(text);
+    let issue : string[] = []; //collect all wrong words in a stringarray
```
Nifdee: You can do this in an easier way:
`const invalidWords = preprocessedText.filter(word => !json.foo.includes(word));`

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    for (const word of preprocessedText) {
+        if (!json.foo.includes(word)) {//word not included in AllowedVocab.json
+            console.log(word + " ist kein erlaubtes Wort. Bitte verwenden Sie ein anderes!"); //for debugging purposes. The user should get a message on screen telling him what's wrong.
+            issue.push(word);
+        }
+    }
+    let allWrongWords :string = "";
+    allWrongWords += issue.join //Join all wrong words to one String
```
Nifdee: This works?!? I assume `issue.join` concatenates without any delimiters - not sure if that's what you want? Generally, even though this might work, you should always call methods in javascript, with parameters. So don't use `issue.join`, but `issue.join(", ")` instead.

Usually, I'd avoid string concatenation using `+=` - it's much nicer to do this instead:
```
const errorMessage = `${invalidWords.join(", ")} sind nicht erlaubte Wörter. Bitte verwenden Sie andere!`;
```
For bonus points, you should only create an error message when there *are* invalid words - so turn it into 
```
const errorMessage = problematicWords.length > 0 ? `${invalidWords.join(", ")} sind nicht...` : undefined;
if (errorMessage !== undefined) {
  console.log(errorMessage);
}
```
and ideally create a separate function `logInvalidWords(words: string[])` for it.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    var textarray = transformTextToStringArray(cleantext);
+    return textarray;
+}
+
+export function checkInput(text: string){
+    let preprocessedText : string[] = preprocessingText(text);
+    let issue : string[] = []; //collect all wrong words in a stringarray
+    for (const word of preprocessedText) {
+        if (!json.foo.includes(word)) {//word not included in AllowedVocab.json
+            console.log(word + " ist kein erlaubtes Wort. Bitte verwenden Sie ein anderes!"); //for debugging purposes. The user should get a message on screen telling him what's wrong.
```
Nifdee: I don't think you need this - you're printing a long error message at the end anyway.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+it("Works for an array containing only the empty String", () =>{
+    const wrongWords = checkInput("");
+    expect(wrongWords).toEqual([""]);
+});
+
+it("Works for an array containing duplicates", () =>{ 
+    const wrongWords = checkInput("bloedsinn bloedsinn");
+    expect(wrongWords).toEqual(["bloedsinn", "bloedsinn"]);
```
Nifdee: Why do we want the output to contain duplicates? Wouldn't it suffice to output `bloedsinn` once?

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import ReactDOM from "react-dom";
+import {checkInput, transformTextToStringArray, eliminatePunctuation} from "./detectWrongSyntax";
+
+
+it("Works for simple text", () => {
+    const wrongWords = checkInput("fu bar bloedsinn"); //fu and bar are indeed included in allowedVocab
+    expect(wrongWords).toEqual(["bloedsinn"]);
+  });
+
+it("Works for an array containing only the empty String", () =>{
```
Nifdee: Generally it's nice if the text explains the behavior in a bit more detail than "it works" - so for this it'd be something like `it("outputs an empty array when input contains only the empty string"...)` or `it("detects the empty string as invalid word"...)` depending on what the expected behavior is. Then a user can simply scan the tests to understand what the functionality is, and doesn't have to look at the test implementation to find out what the expected behavior is.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    const wrongWords = checkInput("");
+    expect(wrongWords).toEqual([""]);
+});
+
+it("Works for an array containing duplicates", () =>{
```
Nifdee: Similarly, here it'd be nicer to use `"preserves duplicates"` or `"returns each invalid word only once"`.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    let allWrongWords :string = "";
+    allWrongWords += issue.join //Join all wrong words to one String
+    allWrongWords += " sind nicht erlaubte Wörter. Bitte verwenden Sie andere!"
+    console.log(allWrongWords); //Tell the User what's wrong
+    return issue;
+}
+
+/**
+ * Delete any unwanted punctuation
```
Nifdee: "Delete any unwanted punctuation" basically says the same as the method name, "eliminate punctuation". A more helpful documentation would be something like "Turns all unwanted punctuation (some unicode characters, as well as specific punctuation signs) into whitespace".

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,79 @@
+import json from "./AllowedVocab.json"
+
+/**
+ * Checks, if all words used, are allowed.
+ * 
+ * 
+ * 
+ */
```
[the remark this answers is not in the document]
Korosensei42: Those empty lines seems to be removed now, however GitHub, doesn't recognize this as outdated

Korosensei42:
I didn't change the tests accordingly, yet.
Will do so tomorrow or saturday.

Korosensei42:
Tests changed. Gonna look into returning the positions of wrong words now.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @type string array.
+ */
+
+export function checkInput(text: string) {
+		const preprocessedText: string[] = preprocessText(text);
+		const invalidWords = preprocessedText.filter(word => !allowedWords.includes(word));
+		// TODO: Make logInvalidWords dependent on this posOfInvalidWords.
+		const posOfInvalidWords = findPositionsOfInvalidWords(text, invalidWords);
+		logInvalidWords(invalidWords);
+		// How to change the Issuemessage?
```
Khazuar: `Issue` is a type. You'll need to create a matching instance here and return it:
```typescript
return {
  message: `Das Wort ${invalidWord} ist unbekannt.`,
  position: {
    startIndex: 42,
    endindex: 42 + invalidWord.length,
  },
};
```

You probably want to return multiple issues at once (in an array), one for each wrong word, including their position.

Korosensei42:
Test for Mapper returns this:
```    Expected value to equal:
      [["test", [0, 3]]]
    Received:
      Map {"test" => [0, 3]}
```

Is this okay for our purposes?

src/vocabularyChecker/Issue.ts
```diff
@@ -0,0 +1,7 @@
+export default interface Issue {
+		position: {
+			fromIndex: number;
+			toIndex: number;
+		};
+		message: string;
+	}
```
Khazuar: Indentation.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @function checkInput then filters out all wrong @param word
+ * from @param text and puts them all into @param allWrongWords,
+ * which is printed in the console log by calling @function logInvalidWords.
+ * This is for debugging purposes and should later be changed to
+ * be displayed to the user. For further postprocessing all wrong
+ * words are collected in a @returns @param invalidWords of
+ * @type string array.
+ */
+
+export function checkInput(text: string) {
```
Khazuar: you can make reading this function easier by specifying the return type explicitly - by adding `...): Whatevertype {`

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @param text
+ * @param invalidWords
+ * Returns a map of invalid words (consisted in @param invalidWords) and their corresponding
+ * positions (starting position and end position) inside of a string.
+ */
+
+export function createMapOfInvalidWords(text: string, invalidWords: string[]) {
+	const positions = new Map();
```
Khazuar: here, you created an implicitly typed map. You can specify the maps key- and value-types by writing `new Map<string, Position>();`.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+	for (const word of invalidWords) {
+		positions.set(word, [text.indexOf(word), text.indexOf(word) + word.length - 1]);
```
Khazuar: While tuple-types `[TYPE1, TYPE2, ... TYPEN]` are great to represent actual "pairs" of things and are sometimes necessary to use in certain APIs (like in `Map`s for their entries), they are not so good at establishing meaning and readability about their components. Especially when the components are of the same type, you gain not much from them.
Enter object-types (`{ prop1: TYPE1, prop2: TYPE2, ... propN: TYPEN }`). These allow you to give your components names and address them by that name, not needing to worry about their order at all. Example:

```typescript
function createPosition(startIndex: number, length: number): { fromIndex: number, toIndex: number } {
  return {
    fromIndex: startIndex,
    toIndex: startIndex + length,
  };
}
```

See how it's much more cleaner now to understand what the separate components mean?

If you plan on using an object type in multiple places, you can define it formally:

```typescript
type Position = { fromIndex: number, toIndex: number };
```

and you can even export those!

An alternative to defining object-types is defining a proper interface:

```typescript
interface Position {
  fromIndex: number;
  toIndex: number;
}
``` 

Since there is no downside to doing that, and interfaces can easily extend on one another, they are usually preferred. It's really the same thing though.

You *can* define structurally identical object-types or interfaces and assign them to another (Typescript uses a structural type system), but if something is used in multiple places and means the same everywhere, it should all refer to the same definition. In those cases, put the interface in its own file, export it from there and import it in all others.

Of course, you can also use full-size classes for the purpose of building data-structures, C#-style:

```
export class Position {
  public readonly fromIndex: number;
  public readonly toIndex: number;

  constructor(fromIndex: number, toIndex: number) {
    this.fromIndex = fromIndex;
    this.toIndex = toIndex; 
  }
}
```

These can also contain getters, setters, methods, static-functions ... it's usually overkill for something so simple though.

But it's all better than tuples.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+ * be displayed to the user. For further postprocessing all wrong
+ * words are collected in a @returns @param invalidWords of
+ * @type string array.
+ */
+
+export function checkInput(text: string) {
+		const preprocessedText: string[] = preprocessText(text);
+		const invalidWords = preprocessedText.filter(word => !allowedWords.includes(word));
+		// TODO: Make logInvalidWords dependent on this posOfInvalidWords.
+		const mapOfInvalidWords = createMapOfInvalidWords(text, invalidWords);
```
Khazuar: You already searched through the text to find your wrong words, now you search through it again? I know that O(n)=O(2n), but still, this is just unnecessary ;) You can just find the positions during your first go.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/**
+ * Collects all positions where there is a whitespace in a string.
+ */
+
+function findSpacePositionsInStringArray(text: string) {
+	let index = 0;
+	const spacePositions = [];
+	while ((index = text.indexOf(" ", index + 1)) > 0) {
```
Khazuar: there is an `includes` method.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,142 @@
...
+export function splitIntoWords(text: string) {
+		text = text.trim();
+		const spacePositions = findSpacePositionsInStringArray(text);
+		let tempPos = 0;
+		const result: string[] = [];
+				// this loop is problematic. Adds a whitespace to the beginning of each word after the second one.
+		spacePositions.forEach(element => {
+				result.push(text.substring(tempPos, element));
+				tempPos = element;
+		});
```
Khazuar: Please do this all with a regex instead that matches words surrounded by word barriers. After that, this function should be a simple loop. And you can even get the index of your findings that way.

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: includes was not a helpful idea. However, I should test the function, if it works correct for texts like "hello,@;)(/ world". Should only result in one space position. Pipeline may be wrong however and return multiple ones, since ,@;)(/ are all replaced with a Whitespace. In that case `trim` beforehand.

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: Instead of Position I use [number, number] as value. The first number representing the first letter of a word and the second number representing the last letter of a word, where word is the key. Position seems to work equivalently though.

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: I now played around with some regular expressions and got /\b(\w*)\w\b/g, but how does that help me to get the positions  of the word barriers numerically?

Korosensei42:
I looked around a bit more on the internet for using regular expressions and with them also the position of words. After tinkering around for a while, I overthought my whole pipeline and basically omitted most of it.
Now I think most of it can be done with Regex and very few codelines. I have not fleshed out this idea very much, but I will try this different ansatz now.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+interface Position { fromIndex: number; toIndex: number; }
+
+/**
+ * @param text
+ * @param invalidWords
+ * Returns a map of invalid words (consisted in @param invalidWords) and their corresponding
+ * positions (starting position and end position) inside of a string.
+ */
+
+export function createMapOfInvalidWords(text: string, invalidWords: string[]): Map<string, [number, number]> {
```
Nifdee: Why don't you use the `Position` interface here? If you return `Map<string, Position>` instead, the comment above the function wouldn't be necessary. Or you could change it to something like `Maps all invalid words to the positions where they occur in text`. 
Why does the map only contain one position for each invalid word, not all occurrences of that word? Shouldn't it rather return something like `Map<string, Position[]>`?

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+	return positions;
+}
+
+/**
+ * @param text
+ * @param invalidWords
+ * Returns message that tells the user which words are not allowed and on which
+ * position those are within @param text.
+ */
+export function createErrorMessages(text: string, invalidWords: string[]): string|undefined {
```
Nifdee: I think this should be called `createErrorMessage` (not plural) since it returns only one `string`. Also, no need to create the `wrongWords` variable here, you can directly `return logInvalidWords(...);`.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+	const wrongWords = logInvalidWords(createMapOfInvalidWords(text, invalidWords));
+	return wrongWords;
+}
+
+/**
+ * @param text
+ * Just a function to bundle the functionalities of @function collectAllInvalidWords and
```
Nifdee: Comments generally shouldn't describe how a function works internally ("first calls X, then calls Y, and returns Z"), but rather describe what the function should be used for / any edge cases to look out for. Usually it's best to choose function names that already sufficiently describe what the function does, so that comments are not necessary, as they are likely to get out of sync when the code they describe gets changed.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * @param text
+ * @returns a Stringarray containing every occurrence of each wrong word in @param text.
+ */
+export function collectAllInvalidWords(text: string): string[] {
+	const temp = text.match(anyWord) || [];
+	const result: string[] = [];
+	temp.forEach(element => {
```
Nifdee: You can still simplify this function:
```
const words = text.match(anyWord) || [];
return words.filter(word => !allowedWords.includes(word));
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+ return result;
+}
+
+/**
+ * @param invalidWords
+ * @returns a Stringarray containing the same elements as @param invalidWords
+ * minus duplicate elements.
+ */
+function removeDuplicates(invalidWords: string[]): string[] {
+	return invalidWords.filter((value, item) => invalidWords.indexOf(value) === item);
```
Nifdee: You can simplify and generalize this (it works for any type `T`, not just `string`, and `Set` is a type that doesn't contain duplicates):
```
function removeDuplicates<T>(values: T[]): T[] {
  return [...new Set(values)];
}
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Just a function to bundle the functionalities of @function collectAllInvalidWords and
+ * @function removeDuplicates.
+ * @returns a Stringarray containing exactly one copy of each wrong word in @param text.
+ */
+export function preprocessText(text: string): string[] {
+	return removeDuplicates(collectAllInvalidWords(text));
+}
+
+/**
+ * @param text
```
Nifdee: You don't need to include `@param foo` in any of your comments - the parameters are contained in the function signature. Comments should only give *additional* information that cannot be conveyed by using types or good function names.

src/vocabularyChecker/detectWrongSyntaxOld.ts
```diff
@@ -0,0 +1,144 @@
+import json from "./AllowedVocab.json";
```
Nifdee: What is this file used for? With `git` you don't have to keep code that is no longer used - if you ever need to go back to it, you can look at the history.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * (input:) @param invalidWords.
+ * Prints out a error message on the console (later to the user),
+ * iff there are invalid Words.
+ */
+
+function logInvalidWords(invalidWords: Map<string, [number, number]>) {
+	return invalidWords.size > 0 ? `${invalidWords.forEach(logMapElement)}` : undefined;
```
Nifdee: I'm surprised this works - `forEach` usually doesn't have a return value; you want to use `map` instead. Also, I think it'd be nicer to explicitly concatenate the results:
```
return invalidWords.size > 0 
  ? invalidWords.map(logMapElement).join("\n")
  : undefined;
```

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: If I understand you correctly I could even remove this whole function, if I just stick to using Set from the start, right?

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,77 @@
...
+	const wrongWords = preprocessText(`This is just a really long String to test some extreme case so
+the next letters will only be some copies of the letter a.
+aaaaaaaaaaaaaaaaaaaaaaaaaaa`);
+	expect(wrongWords).toEqual(["This", "is", "just", "a", "really", "long", "String", "to", "test",
+	"some", "extreme", "case", "so", "the", "next", "letters", "will", "only", "be", "copies",
+		"of", "letter", "aaaaaaaaaaaaaaaaaaaaaaaaaaa"]);
+});
+
+it("Recognizes the words correctly", () => {
+		const result = preprocessText("Hallo Welt, fu. foo, bar");
```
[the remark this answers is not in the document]
Korosensei42: I think it should be treated as hello and world.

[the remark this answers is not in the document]
Korosensei42: Actually it get's treated as `hello.` and `world`. In that case I might need to overthink my regular expression.

src/vocabularyChecker/detectWrongSyntax.ts
Khazuar: If sets are all you need then yes, go for it. 

Don't overestimate sets in general though; they use `!(a !== b)` to compare elements stored within them, *not a deep comparison*. Keep that in mind before storing objects or arrays in them.

If you decide to go with @Nifdee s suggestion for a utility method, I would implement it as 
```typescript
export function uniqueValues<T>(values: Iterable<T>): Iterable<T> {
  return new Set(values);
}
```
and put it into it's own file (or a file that contains all the `Iterable`-functions we'll come up with.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import {collectAllInvalidWords, createErrorMessages} from "./detectWrongSyntax";
+import {createMapOfInvalidWords, logMapElement, preprocessText} from "./detectWrongSyntax";
+it("Returns the correct word for a normal text", () => {
```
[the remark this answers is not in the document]
Khazuar: Indentation. Please indent nested codeblocks with a tab.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,91 @@
...
+	position: {fromIndex: 0, toIndex: text.length}});
+});
+}),
+
+describe("createMapOfInvalidWords", () => {
+it("Creates a correct Map", () => {
+	const issues = createMapOfInvalidWords(`test`, [`test`]);
+	expect(issues.entries()).toEqual(["test", [0, 3]]);
+});
+}),
```
Khazuar: Indentation as well.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+describe("createErrorMessages", () => {
+it("Creates a correct error message", () => {
+	const issues = createErrorMessages(`test1`, [`test1`]);
+	expect(issues).toEqual("test1");
+});
```
Khazuar: Indentation.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Regular expression that looks for any and all words. A word is defined as some sequence of
+ * characters (\w*S) betweeen wordbarriers (\b)
+ */
+const anyWord = new RegExp(/\b(\w*\S)\b/g);
+/**
+ * Predefined dictionary containing all allowed Words
+ */
+const allowedWords = json;
+interface Position { fromIndex: number; toIndex: number; }
```
Khazuar: Export please.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param text
+ * @param invalidWords
+ * Returns a map of invalid words (consisted in @param invalidWords) and their corresponding
+ * positions (starting position and end position) inside of a string.
+ */
+
+export function createMapOfInvalidWords(text: string, invalidWords: string[]): Map<string, Position[]> {
+	const positions = new Map<string, Position[]>();
+	for (const word of invalidWords) {
+		positions.set(word,  text.indexOf(word), (text.indexOf(word) + word.length - 1);
```
Khazuar: Syntax here is broken. Has this code been tested?

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,91 @@
...
+export function getInvalidWords(text: string): string[] {
+	return removeDuplicates(collectAllInvalidWords(text));
+}
+
+/**
+ * @param text
+ * @returns a Stringarray containing every occurrence of each wrong word in @param text.
+ */
+export function collectAllInvalidWords(text: string): string[] {
+	const words = text.match(anyWord) || [];
```
Khazuar: `[...text.matchAll(anyword)]` will give you the index of every match as well. You can then return the invalid words and their positions here directly.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,91 @@
...
+	return invalidWords.size > 0 ? `${invalidWords.forEach(logMapElement)}` : undefined;
+}
+
+/**
+ * @param value consisting of value[0], the fromIndex and value[1], the toIndex
+ * @param key
+ * returns an Issue-like object, telling the user which word is not allowed and on which position it is.
+ */
+
+export function logMapElement(value: [number, number], key: string): Issue {
```
Khazuar: `key` and `value` are really bad names for variables. Also: Replace the tuple type with `Position`.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,91 @@
...
+	return {message: `${key} an Stelle ${value} ist ein unerlaubtes Wort! \n`,
+		position: {
+			fromIndex: value[0],
+			toIndex: value[1],
+		},
+		};
```
Khazuar: Indentation.

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: When trying `const words = text.matchAll(anyWord) || [];` I get an Syntaxerror followed by this errormessage:

Property 'matchAll' does not exist on type 'string'

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: In this function I don't get Syntaxerrors. However, you are right with respect to this function

```export function logMapElement(value: Position[], key: string): Issue {
	return {message: `${key} an Stelle ${value} ist ein unerlaubtes Wort! \n`,
		position: {
			fromIndex: value[0],
			toIndex: value[1],
			},
		};
	}
```

Here I get the error

Type 'Position' is not assignable to type 'number'.

Also my attempts to fix this did not work. I will try something different

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -4,88 +4,88 @@ import {collectAllInvalidWords, createErrorMessages} from "./detectWrongSyntax";
...
-	const text = `test Test`;
-	const issues = logMapElement([0, text.length], text);
-	expect(issues).toEqual({message: text + " an Stelle 0," + text.length + " ist ein unerlaubtes Wort! \n",
-	position: {fromIndex: 0, toIndex: text.length}});
-});
+	it("Logs a String containing multiple words correctly", () => {
+		const text = `test Test`;
+		const issues = logMapElement([0, text.length], text);
+		expect(issues).toEqual({message: text + " an Stelle 0," + text.length + " ist ein unerlaubtes Wort! \n",
+		position: {fromIndex: 0, toIndex: text.length}});
```
Khazuar: Please separate object literals onto multiple lines, so it's easier to read:

```typescript
expect(issues).toEqual({
  message: text + " an Stelle 0," + text.length + " ist ein unerlaubtes Wort! \n",
  position: {
    fromIndex: 0,
    toIndex: text.length,
  },
});
```

Please notice the `,` after every property (including the last one). This is so git (which diffs line-by-line) doesn't count added commas in the line before as changed lines. Common convention.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -1,15 +1,16 @@
+import { Interface } from "readline";
```
Khazuar: I don't think you want that to be in here.

src/vocabularyChecker/detectWrongSyntax.ts
TimothyGillespie: Does `const words = anyWord.matchAll(text)` work?

Korosensei42: In that case I get the same error message, only now "string" is replaced by "RegExp"

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: Solved it by a new Issue-based idea

src/vocabularyChecker/detectWrongSyntax.ts
Khazuar: Reason is that `matchAll` is pretty new: https://stackoverflow.com/questions/55691536/does-node-js-support-the-string-matchall-method/55691566.
[Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#Browser_compatibility) it actually states that node.js (the engine running the tests) does not support it yet. 
We **could** add a polyfill to enable it, but I doubt it's worth the effort for now. I you want to spend a few minutes on that (it probably means installing / updating a few npm packages) feel free to do so, but if you bite your teeth out on it for more than 10min, leave it be I'd say.

Korosensei42: I found this: https://babeljs.io/docs/en/babel-polyfill. Is this here all I need to install?

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: For now, removeDuplicates works appropriately. I will see how it works with Sets, though.

Korosensei42:
Did some additional changes. I will only try the thing with the Set now, but I think my code does what it's supposed to do now.

src/vocabularyChecker/Issue.ts
Khazuar: Please merge with master and use the new unified issue interface instead.

src/vocabularyChecker/detectWrongSyntax.ts
Khazuar: maybe. Maybe `core-js` would be fine as well. I'll leave that to you to experiment on in a separate branch :)

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return result;
+}
+
+export function logMultipleOccurences(word: string, position: Position[]): Issue[] {
+	const result = [];
+	for (const pos of position) {
+		result.push(logSingleWord(word, pos));
+	}
+	return result;
```
Khazuar:
```typescript
return position.map(pos => logSingleWord(word, pos));
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export function logSingleWord(word: string, position: Position): Issue {
+	return {message: `${word} von Stelle ${position.fromIndex} bis ${position.toIndex} ist ein unerlaubtes Wort! \n`,
```
Khazuar: New line before `message:` !

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * The next three functions print a Message for every occurence of every invalid word.
+ */
+
+export function logMultipleWords(words: string[], positions: Position[]): Issue[][] {
+	const result = [];
+	for (const word of words) {
+		result.push(logMultipleOccurences(word, positions));
+	}
+	return result;
```
Khazuar:
```typescript
return words.map(word => logMultipleOccurences(word, positions));
```

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+	return collectInvalidWordsInIssues(text, collectAllInvalidWords(text));
+}
+
+export function collectInvalidWordsInIssues(text: string, invalidWords: string[]): Issue[] {
+	const issues = [];
+	const positions: Position[] = [];
+	for (const word of invalidWords) {
+		const temp: Position = {fromIndex: text.indexOf(word), toIndex: (text.indexOf(word) + word.length)};
+		positions.push(temp);
+		issues.push({message : word, position: temp});
```
Khazuar: Nit: No whitespace before `:`

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export function getAllIssues(text: string): Issue[] {
+	return collectInvalidWordsInIssues(text, collectAllInvalidWords(text));
+}
+
+export function collectInvalidWordsInIssues(text: string, invalidWords: string[]): Issue[] {
+	const issues = [];
+	const positions: Position[] = [];
+	for (const word of invalidWords) {
+		const temp: Position = {fromIndex: text.indexOf(word), toIndex: (text.indexOf(word) + word.length)};
```
Khazuar: Put that object literal on multiple lines please.

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Regular expression that looks for any and all words. A word is defined as some sequence of
+ * characters (\w+) betweeen wordbarriers (\b)
+ */
+const anyWord = new RegExp(/\b\w+\b/g);
+/**
+ * Predefined dictionary containing all allowed Words
+ */
+const allowedWords = json;
+export interface Position { fromIndex: number; toIndex: number; }
```
Khazuar: Reuse `Position` from the master branch.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,138 @@
...
+	it("Detects a String of word1.word2 as two words", () => {
+		const issues = getAllIssues("Te.st");
+		expect(issues).toEqual(
+			[{message: "Te", position: {fromIndex: 0, toIndex: 2}}, {message: "st", position: {fromIndex: 3, toIndex: 5}}]);
+	});
+
+	it("Detects a String of word1,;:<>=word2 as two words", () => {
+		const issues = getAllIssues("Te,;<>=st");
+		expect(issues).toEqual(
+			[{message: "Te", position: {fromIndex: 0, toIndex: 2}}, {message: "st", position: {fromIndex: 7, toIndex: 9}}]);
```
Khazuar: Please leave the opening `[` on the same line as the opening `(` and put the `]);` on its own line. Add a linebreak after each array element and a comma after the last one.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -0,0 +1,138 @@
...
+describe("logSingleWord", () => {
+	it("Logs a ordinary word correctly", () => {
+		const position: Position = {fromIndex: 0, toIndex: 4};
+		const issues = logSingleWord("Hello", position);
+		expect(issues).toEqual(
+			{message: `${"Hello"} von Stelle ${position.fromIndex} bis ${position.toIndex} ist ein unerlaubtes Wort! \n`,
+		position: {
+			fromIndex: position.fromIndex,
+			toIndex: position.toIndex,
+			}});
```
Khazuar: Indentation. `position` should be indented once more, the closing `}` should be indented one less.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -78,61 +83,150 @@ describe("logSingleWord", () => {
...
 	it("Logs a word consisting only of a whitespace correctly", () => {
 		const position: Position = {fromIndex: 0, toIndex: 1};
 		const issues = logSingleWord(" ", position);
 		expect(issues).toEqual(
-			{message: `${" "} von Stelle ${position.fromIndex} bis ${position.toIndex} ist ein unerlaubtes Wort! \n`,
-		position: {
-			fromIndex: position.fromIndex,
-			toIndex: position.toIndex,
-			}});
+			{
```
Khazuar: Nit: Put the opening bracket on the line before.

Khazuar:
Just the indentation, otherwise good to merge.

PR #11, opened 2019-05-02
Timothy gillespie/tau prolog

TimothyGillespie:
The rough basic functions is working as intended. The idea is to group feedback in nested Arrays. Example:

Prolog-Program:
likes(lisa, bob, frank).
likes(lisa, frank, bob).

Query:
likes(lisa, X, Y).

We would expect Prolog to say that the variables could be set the following way:
X = bob,
Y = frank;

or alternatively:
X = frank,
Y = bob;

The return would end in SWI with
X = frank,
Y = bob.

in Tau-Prolog there will always be an extra "false."

The output will model Tau-Prolog since it is easier to implement and having a strict SWI-Prolog consistent output isn't beneficial.

So our return for the above case would be:
[["bob", "frank"], ["frank", "bob"], ["false"]]

More examples can be found in the test file.


src/util/prolog/prologSession.ts
```diff
@@ -0,0 +1,90 @@
...
+		constructor(program: string, limit?: number) {
+				this.session = new type.Session(limit);
+				this.session.consult(program);
+		}
+
+		// Options
+		// 0 means return all values (boolean and variable values)
+		// 1 means only return boolean returns
+		// 2 means only return variable values
+		public async executeQuery(query: string, option: number = 0): Promise<string[][]> {
```
Khazuar: ## Multiple pieces of advise here
### Types
Typescript supports both enums and union-types. Both are more expressive than bare `number`s. 

Example for union type:
```typescript
public async executeQuery(query: string, filter?: "booleans" | "variables"): ...
```
Of course you can define the union type separately as well:
```typescript
(export) type FilterSetting = "booleans" | "variables";
...
public async executeQuery(query: string, filter?: FilterSetting): ...
```

Example for enum type:
```typescript
(export) enum FilterSetting {
  None,
  Boolean,
  Variables,
}
...
public async executeQuery(query: string, filter: FilterSetting): ...
```

### Style
When a function does different things based on a parameter (or even returns different types!) it's good to split the function up or at least make it private and offer public specific wrappers to make it easier to use.

src/util/prolog/prologSession.ts
```diff
@@ -0,0 +1,90 @@
...
+					results.splice(ct, 1);
+					results.splice(cf, 1);
+				}
+				results.indexOf(["false"]);
+			}
+			*/
+
+		}
+
+		private formatNestedArray(source: string, goal: string[][]): void {
```
Khazuar: This method could just return an "inner" array (type: `string[]`). It would'n need the reference to the "outer" array. It would be used like: `results.push(this.formatNestedArray(result))`. Consider renaming it accordingy.

src/util/prolog/prologSession.ts
```diff
@@ -0,0 +1,90 @@
...
+			// There is probably a more elegant way to fetch these cases
+			if (source.includes("true")) { goal[goal.length - 1][0] = "true"; return; }
+			if (source.includes("false")) { goal[goal.length - 1][0] = "false"; return; }
+
+		 const pattern: any = /= (\w*|\[.*\])?(,| ;|.)/g;
+		 const components: any = source.match(pattern);
+
+		 components.forEach((e: any) => {
+				// Throws out the first two chars (= and a blank) and the last char (either , ; or .)
+				const toAdd: string = e.substring(2, e.length - 1).trim();
```
Khazuar: This is to remove the `X =`?

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,246 @@
...
+			grandfather(X, Z):-
+			grandparent(X, Z),
+			male(X).
+
+			grandmother(X, Z):-
+			grandparent(X, Z),
+			female(X).
+				`, 1000);
+
+		test.each`
```
Khazuar: if you use `test.each`, please make sure you format the text so it actually increases readability.

src/util/nestedArray/nestedArray.ts
```diff
@@ -0,0 +1,22 @@
+export function nestedIndexOf(hay: any[][], needle: any[]): number {
+	let i = 0;
+
+	hay.forEach((e: any) => {
+		if (e === needle { return i; }
+		i = i + 1;
+	});
+
+	return -1;
+}
```
Khazuar: not quite sure what this method is doing differently to `Array.includes`.

src/util/nestedArray/nestedArray.ts
```diff
@@ -0,0 +1,22 @@
+export function nestedIndexOf(hay: any[][], needle: any[]): number {
+	let i = 0;
+
+	hay.forEach((e: any) => {
+		if (e === needle { return i; }
```
Khazuar: you don't close the opening parenthesis of the `if` here.

src/util/nestedArray/nestedArray.ts
```diff
@@ -0,0 +1,22 @@
+export function nestedIndexOf(hay: any[][], needle: any[]): number {
+	let i = 0;
+
+	hay.forEach((e: any) => {
```
Khazuar: you can add a second parameter to this function that will be the index: `(item: any, index: number)`.

src/util/nestedArray/nestedArray.ts
```diff
@@ -0,0 +1,22 @@
+export function nestedIndexOf(hay: any[][], needle: any[]): number {
```
Khazuar: Such methods should use generics:
```typescript
export function nestedIndexOf<T>(hay: T[][], needle: T[]): number { ...
```

src/util/nestedArray/nestedArray.test.ts
```diff
@@ -0,0 +1,15 @@
+import {nestedIndexOf} from "./nestedArray";
+
+test.each`
```
Khazuar: Please improve the formatting here as well, so `test.each` actually increases readability.

src/util/prolog/prologSession.ts
TimothyGillespie: Partially. The `X ` is filtered out on line 81. It would also filter out variables which are not only one letter.
Line 85 only filters out the `=` and the final symbol which is either a semikolon, comma or dot.

src/util/nestedArray/nestedArray.ts
TimothyGillespie: It is supposed to handle and return the index of an array in a nested array since javascript natively doesn't handle nested arrays as expected. I will try the other solutions you have mentioned in the call with the lodash library and the filter methode.

src/util/prolog/prolog.test.ts
TimothyGillespie: I think it is more readibly since it get rid of the repeated structurs which only clutter the code. I can at least find my input and expected output easier this way. It also looks different in the editor. The comments make it less readable on github.

This also has another advantage. If I change the way the results are accessed I only have to change it one location for all tests and not in every test method.

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,285 @@
...
+				expect(result).toStrictEqual([["bob", "karl", "friedrich"], ["bob", "friedrich", "karl"], ["false"]]);
+
+		});
+
+		test("Four Variables Two Results", async () => {
+				const session = new PrologSession(`
+					likes(lisa, bob, karl, friedrich, dieter).
+					likes(lisa, bob, karl, dieter, friedrich).
+					`, 1000);
+				const result = (await session.executeQuery("likes(lisa, X, Y, Z, A).")).getResults();
```
Khazuar: Ir reads a little bit strange that you call it `result`, even though the method is called `getResults` and it's actually an array.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,54 @@
...
+		if (this.results.length >= 1 ) { return this.results; }
+
+		this.rawResults.forEach((e: any) => {
+			this.results.push(this.rawResultsToInnerArray(e));
+		});
+
+		return this.results;
+	}
+
+	public filterBooleanResults(): string[][] {
```
Khazuar: Since this method filters for boolean results, why don't you return `boolean`s?

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,54 @@
...
+		const copy: string[][] = this.getResults();
+		return copy.filter((e: any) => {
+			return e[0] === "true" || e[0] === "false";
```
Khazuar: what about checking the length of `e`? Also, you can write this as `.filter((e: any) => e[0] === "true" ...)`

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,54 @@
...
+		} );
+	}
+
+	public filterVariableResults(): string[][] {
```
Khazuar: What about assigning the results to the used variables somehow? Like returning a map or maps?

src/util/prolog/prologResult.ts
```diff
@@ -1,6 +1,9 @@
 export class PrologResult {
-	private readonly results: string[][];
 	private readonly rawResults: string[];
+	private readonly results: string[][];
+
+	// First Group will be the variable value. Doesn't fetch the variable names
+	private readonly prologAnswerPattern: string = " = (\\w*|\\[.*\\])?(,| ;|.)";
```
Khazuar: No need to store this in every instance of the class, since it's effectively a constant. You can put it into a not-exported constant outside the class:
```
const prologAnswerPattern = " = (\\w*|\\[.*\\])?(,| ;|.)";
```
You might even want to create a regex literal:
```
const prologAnswerPattern = / = (\w*|\[.*\])?(,| ;|.)/;
```
This saves a few backslashes (which can be donated to Microsoft, their OS uses a lot of them ^^).

Khazuar: Edit: I see you need to add the variable name to it. No regex literal then.

src/util/prolog/prologResult.ts
```diff
@@ -17,11 +20,17 @@ export class PrologResult {
 		return this.results;
 	}
 
+	public getResultsFor(variable: string): string[] {
+		// Group 1 will be the content of the variable independent of the variable length.
+	 const pattern: any = new RegExp(variable + this.prologAnswerPattern, "g");
```
Khazuar: Why does this need to be `any`?

src/util/prolog/prologResult.ts
TimothyGillespie: Exactly! I need to keep it as a string to enhance it later. But it sure makes sense to save it outside of the class.

src/util/prolog/prologResult.ts
TimothyGillespie: Seems like I misunderstood typescript types. It should rather be `RegExp` right?

src/util/prolog/prologResult.ts
TimothyGillespie: I implemented it that way now. Made some of the other functions superfluous.

src/util/prolog/prologResult.ts
TimothyGillespie: Will be solved by utilizing `RegExp` groups now in another function.

src/util/prolog/prologResult.ts
TimothyGillespie: The idea was to keep it consistent with all the other outputs, but thinking about it now it didn't make a lot of sense.

src/util/prolog/prologResult.ts
Khazuar: Yes, the type would be `RegExp`. Since you just assign `new RegExp` to this constant variable though, an explicit type is superfluous and doesn't increase typesafety nor readability. I'd just write `const pattern = new Regexp(...);`

desktop.ini
```diff
@@ -0,0 +1,4 @@
+[ViewState]
```
Nifdee: I don't think this file is supposed to be committed?

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+	const groupContent: string[] = [];
+	let aux: string[];
+
+	while ((aux = pattern.exec(basis)) !== null) {
+		groupContent.push(aux[group]);
+	}
+
+	return groupContent;
+}
+	/* With the usage of maps not necessary
```
Nifdee: You don't need to keep old code in here - just delete it, if anyone ever needs it again, it's still available in the `git` history.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
+import * as ld from "lodash";
```
Nifdee: More conventionally, `lodash` is imported as `_`. But I prefer explicitly importing only those methods that you really use.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+
+		results.set("Boolean", new Map());
+		i = 0;
+
+		this.getBooleans().forEach((e: boolean) => {
+			results.get("Boolean").set(i, e);
+			i++;
+		});
+
+		return results;
```
Nifdee: Shouldn't you also set `this.results = results` (or copy all the values from `result` into `this.result`)? Otherwise `this.results` will always be an empty map.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+	private readonly rawResults: string[];
+	private readonly results: Map<string, Map<number, string|boolean>>;
+
+	constructor(rawResults: string[]) {
+		this.rawResults = rawResults;
+		this.results = new Map();
+	}
+
+	public getResults(): Map<string, Map<number, string|boolean>> {
+		if (this.results.size !== 0) { return this.results; }
```
Nifdee: But what if `results` is supposed to be empty? I think it'd be nicer to change the type of `results` to `Map<...> | undefined`, and check if it is `undefined` here.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
+import * as ld from "lodash";
+
+// First Group will be the variable value. Doesn't fetch the variable names
+const prologAnswerPattern: string = " = (\\w*|\\[.*\\])?(,| ;|.)";
+
+export class PrologResult {
+	private readonly rawResults: string[];
+	private readonly results: Map<string, Map<number, string|boolean>>;
```
Nifdee: Does this have to be a cryptic type like ` Map<string, Map<number, string | boolean>>`? In general, unless you really need the `Map` functionality (i.e. quick access by key), it's better to create interfaces because they describe much better what the data means. E.g. your `Boolean` could be a property (that probably shouldn't be called "boolean" - what does it mean?)

src/types/tau-prolog/index.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "tau-prolog/modules/core" {
+    export const type: any;
```
Nifdee: It would be nice to eventually improve these typings; e.g. to add typings for all properties that you directly access. Here you could already use something like the following (or more specific, if you know more about `Session` and the parameter type / return type of `format_answer`):
```
export const type: {
  Session: any;
};
export const format_answer: (x: any) => any;
```

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,296 @@
+
+import { PrologSession } from "./prologSession";
+import * as program from "./testPrograms";
+
+// Expected results are consistent with SWI-Prolog behavior except that they usually have an extra "false."
```
Nifdee: What does the `"false"` mean? Why is it there? I didn't quite understand that from the PR description either.

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+describe("Tau-Prolog Behavior, PrologSession.executeQuery() & PrologResult.getResultArray()", () => {
```
Nifdee: Why is there one `describe` for three different things? Usually if you want to nest `describe` blocks it should be something like 
```
describe("my super nice module", () => {
  describe("some function", () => {
    it("does great things", () => {...});
  });
});
```
Arbitrarily grouping things together doesn't add much value. I've also never needed more than one level of `describe` - usually there aren't that many different functions to test in one file, so it's sufficient to use one `describe` block for each function, and have one `it` / `test` for each test case.

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+		test("Cut Operator Maximum Function", async () => {
+
+				const session = new PrologSession("max(X,Y,Y)  :-  X  =<  Y, !. max(X, Y, X).", 1000);
+				const results = (await session.executeQuery("max(3, 7, X).")).getResultArray();
+				expect(results).toStrictEqual([["7"], ["false"]]);
+
+		});
+
+		test("Cut Operator Endless Loop", async () => {
```
Nifdee: Hm, could you use sentences as a description? Usually these strings are something like `"returns an empty string if the input is invalid"` or `"returns each word only once"` or `"throws an exception if reader is not initialized"` - basically describe the expected behavior in a readable way.

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+		});
+
+		test("Truth Value (True) Perfomance", async () => {
+				const session = new PrologSession("likes(bob, lisa).", 1000);
+				const results = (await session.executeQuery("likes(bob, lisa).")).getResultArray();
+				expect(results).toStrictEqual([["true"], ["false"]]);
+		});
+
+		/* No longer needed with the introduction of maps.
```
Nifdee: There shouldn't be any commented-out code anywhere - if you no longer need it, just delete it. If you ever want to look at this code again it's always there for you in the `git` history.

src/util/prolog/prolog.test.ts
Nifdee: The indentation is a bit weird, maybe you could move the inputs one more level to the left?

src/util/prolog/prologResult.ts
Nifdee: What is the `number` used for? It looks like you just use that to keep track of the order in which the `string | boolean` values are found? Why do you need a `Map<number, string | boolean>` instead of a simple array `Array<string | boolean>`?

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+
+	private getResultsFor(variable: string): string[] {
+		// Group 1 will be the content of the variable independent of the variable length.
+	 const pattern: any = new RegExp(variable + prologAnswerPattern, "g");
+	 return this.regexGroupToArray(this.rawResults.toString(), pattern, 1);
+	}
+
+	private rawResultsToInnerArray(source: string): string[] {
+
+	if (source.includes("true")) { return ["true"]; }
```
Nifdee: Not sure what valid values for `source` are, but couldn't you also have something like `Gertrued` (misspelled name :D)? In that case I don't think you'd want to return `["true"]`?

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+	if (source.includes("false")) { return["false"]; }
+
+	const pattern: any = new RegExp(prologAnswerPattern, "g");
```
Nifdee: Why do you use `any` as a type for `pattern`? Also, you can use the same instance of the regex repeatedly, so just store a `const prologAnswerRegex = new RegExp(prologAnswerPattern, "g");` in the class and always use that.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+	return this.regexGroupToArray(source, pattern, 1);
+	}
+
+	private getVariables(): Set<string> {
+		const variables = new Set();
+
+		// Second group contains the variable names
+		const pattern: RegExp = /(^| )([A-Z]\w*)/g;
```
Nifdee: For all of these you can also just store constants at the top of the file and reuse them.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+		const rawVariables: string[] = this.regexGroupToArray(this.rawResults.toString(), pattern, 2);
+
+		rawVariables.forEach((e: string) => {
```
Nifdee: Instead, you can simply use `return new Set(rawVariables);`.

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+			variables.add(e);
+		});
+	 return variables;
+	}
+
+	private getBooleans(): boolean[] {
+		const pattern: RegExp = /(true|false)(,| ;|.)/g;
+		let results: any[] = this.regexGroupToArray(this.rawResults.toString(), pattern, 1);
```
Nifdee: Why is this `any`? Shouldn't it be `string`? Also, instead of making `results` mutable, simply return the mapping results directly:
```
return results.map((e: string) => e === "true");
```

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+		results = results.map((e: string) => {
+			return (e === "true") ? true : false;
+		});
+
+		return results;
+}
+
+private regexGroupToArray(basis: string, pattern: any, group: number): string[] {
+	const groupContent: string[] = [];
+	let aux: string[];
```
Nifdee: Not a fan of `aux` as a name - what does `aux` store? Maybe call it `matchGroups` instead?

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,120 @@
...
+export class PrologResult {
+	private readonly rawResults: string[];
+	private readonly results: Map<string, Map<number, string|boolean>>;
+
+	constructor(rawResults: string[]) {
+		this.rawResults = rawResults;
+		this.results = new Map();
+	}
+
+	public getResults(): Map<string, Map<number, string|boolean>> {
```
Nifdee: What is this function supposed to do? Feels like there's a simpler way to achieve that, but I don't really understand what this function is needed for.

src/types/tau-prolog/index.d.ts
Khazuar:
```typescript
export namespace type {
  export class Session;
}
export function format_answer(x: any): string;
```

I think.

src/util/prolog/prolog.test.ts
```diff
@@ -1,296 +1,128 @@
...
-		test("One Variable One Result", async () => {
-
-				const session = new PrologSession("likes(lisa, bob).", 1000);
-				const results = (await session.executeQuery("likes(lisa, X).")).getResultArray();
-				expect(results).toStrictEqual([["bob"], ["false"]]);
+		test("Evaluates Equalities Correctly (True Queries)", async () => {
+				const session = programs.emptyCode;
+				const results = (await session.executeQuery("2 = 2.")).getRawResults();
 
+				expect(results).toStrictEqual(["true ;", "false."]);
```
Khazuar: I wonder: Does prolog always use `;` to separate individual results and `.` at the very end? In that case, they don't offer any more information in the output and we could omit them (returning/expecting `["true", "false"]` instead).

Khazuar: But then again, this method is called `rawResults`. So it's probably fine like that.

src/util/prolog/prologResult.ts
```diff
@@ -1,42 +1,37 @@
...
+	}
 
-		const results = new Map();
-		let i: number;
+	/**
+		* Generates (if needed) and returns the results.
+		*
+		*
+		*/
+	public getResults(): Map<string, Array<string | boolean>> {
```
Khazuar: Nit: To protect the encapsulation of `PrologResult`, you should either change the return type of `getResults` to `ReadonlyMap<...>` or create a deep copy before returning!

src/util/prolog/prologResult.ts
```diff
@@ -51,70 +46,45 @@ export class PrologResult {
...
 
 	private getBooleans(): boolean[] {
-		const pattern: RegExp = /(true|false)(,| ;|.)/g;
-		let results: any[] = this.regexGroupToArray(this.rawResults.toString(), pattern, 1);
-		results = results.map((e: string) => {
-			return (e === "true") ? true : false;
-		});
+		const pattern = /(true|false)(,| ;|.)/g;
+		const results: string[] = this.regexGroupToArray(this.rawResults.toString(), pattern, 1);
+		const booleanResults: boolean[] = results.map((e: string) => e === "true");
```
Khazuar: Nit: I don't think you strictly **need** to give an explicit type to bound variables in situations like this - but if it suits you, that is fine by me.

src/util/prolog/prologResult.ts
```diff
@@ -51,70 +46,45 @@ export class PrologResult {
...
-	return groupContent;
-}
-	/* With the usage of maps not necessary
-
-	public filterBooleanResults(): boolean[] {
-		let copy: string[][] = this.getResultArray();
-		copy = copy.filter((e: any) => e.length === 1 && (e[0] === "true" || e[0] === "false"));
-
-		const flatCopy: any[] = ld.flattenDeep(copy);
+	private regexGroupToArray(basis: string, pattern: any, group: number): string[] {
```
Khazuar: why is `pattern` of type `any`? Shouldn't it rather be a `RegExp`?

src/util/prolog/testPrograms.ts
```diff
@@ -43,4 +43,23 @@ export const familySession = new PrologSession(`
   grandmother(X, Z):-
   grandparent(X, Z),
   female(X).
-    `, 1000);
+  `, 1000);
+
+export const max = new PrologSession(`
+  max(X,Y,Y)  :-  X  =<  Y, !.
+  max(X, Y, X).
+  `, 1000);
```
Khazuar: How comes you're always setting the step limit to 1000? Do you strictly need it or could you omit it? If it's just some "large number that will surely never get hit and I only use it so my tests will always terminate", please create a constant for that ;)

Khazuar:
@TimothyGillespie This is overall very good already. Once you're happy with the code, feel free to hit the "Ready for review" button below to signal a final review.

src/util/prolog/prologResult.ts
TimothyGillespie: The string to construct the regex is only used here because it has to be modified later. If store the thing as a regex itself I cannot reuse it.

src/util/prolog/prologResult.ts
TimothyGillespie: This function is supposed to return the cleaned up results or (if they haven't been created yet) it will create them from the rawResults. Using Arrays and implementing your other hints it is a lot simpler now.

src/types/tau-prolog/index.d.ts
```diff
@@ -0,0 +1,10 @@
+export namespace type {
```
TimothyGillespie: I am not quiet sure if it is properly done like this. There is now an linter error on the import in prologSession.

Khazuar: You could import `type.Session` there directly, can't you?

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,128 @@
+
```
Khazuar: Nit: Remove empty line

src/util/prolog/prolog.test.ts
```diff
@@ -0,0 +1,128 @@
...
+		});
+
+});
+
+describe("PrologResult.getResults()", () => {
+	test('The evenNumberFacts Program returns the correct Map for "even(x)."', async () => {
+		const session = programs.evenNumbersFacts;
+		const results = (await session.executeQuery("even(X).")).getResults();
+
+		const expectedResults = new Map().set("X", ["2", "4", "6", "8"]).set("booleanAnswers", [false]);
```
Khazuar: please use either 
```typescript
new Map<string, string[]>()
    .set(..., [...])
    ...
    .set(..., [...]);
```

or

```typescript
new Map([
  ["X", [...]],
  ["Y", [...]],
  ...
  [..., ...],
]);
```

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,92 @@
...
+			results.push(this.rawResultsToInnerArray(e));
+		});
+
+		return results;
+	}
+
+	private getResultsFor(variable: string): string[] {
+		// Group 1 will be the content of the variable independent of the variable length.
+	 const pattern = new RegExp(variable + variableAnswerPattern, "g");
+	 return this.regexGroupToArray(this.rawResults.toString(), pattern, 1);
```
Khazuar: Nit: Indentation

src/util/prolog/prologResult.ts
```diff
@@ -0,0 +1,92 @@
...
+
+	private rawResultsToInnerArray(source: string): string[] {
+
+	// Returns ["true"] or ["false"] in case of boolean answers
+	const booleanAnswer = booleanAnswerRegExp.exec(source);
+	if (booleanAnswer !== null) { return [booleanAnswer[1]]; }
+
+	const pattern = new RegExp(variableAnswerPattern, "g");
+	return this.regexGroupToArray(source, pattern, 1);
+	}
```
Khazuar: Indentation.

src/types/tau-prolog/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'tau-prolog/modules/core'{
```
TimothyGillespie: I think this is now good. I couldn't import types.session directly it seems.

PR #12, opened 2019-05-06
Ronja kaehne/text formatter

RonjaKaehne:


src/text_formatter/allowed_Expression_Detectors.ts
```diff
@@ -0,0 +1,60 @@
+// --------------------------------------
+// @author: Ronja K.
+// ---------------------------------------
+
+
+// ---------CurrentExpressionDetector------------
+// ---------every element have their own detect-function--------
+export function detectExpressionElements(input: string) {
```
Khazuar: You can add documentation to your code using the JSDoc-Syntax: 
```
/** This does amazing things 
  */
function superDuper() { ... }
```

src/text_formatter/allowed_Expression_Detectors.ts
```diff
@@ -0,0 +1,60 @@
...
+	input = detectBracketLeft(input);
+	input = detectBracketRight(input);
+	input = detectEquivalence(input);
+	input = detectImplication(input);
+	input = detectImplication(input);
+	input = detectNegation(input);
+	input = detectConjunction(input);
+	input = detectDisjunction(input);
+	return input;
```
Khazuar: You are returning a string here, with symbols replaced by tokens, separated by spaces, right? This is not an especially handy return-type. Consider returning an `Array<Token>` instead, with `type Token = "bracketLeft" | "bracketRight" | "equivalence" |...` and specify that return type explicitly on the function!

src/text_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import ReactDOM from "react-dom";
+
+// --------------------------------------
+// @author: Ronja K.
+// ---------------------------------------
+// IMPORT FUNCTIONS
+//import { detectExpressionElements, textToListFormat, sentenceIntoWordList, expressionFormatter, removeWhiteSpaces, replaceInputCaractersToReadablePrologCharacter } from "./text_formatter";
+import { textToListFormat, sentenceIntoWordList, replaceInputCaractersToReadablePrologCharacter, removeWhiteSpaces, checkAmountOfBrackets } from "./text_formatter";
+import { detectExpressionElements } from "./allowed_Expression_Detectors";
```
Khazuar: Why is your author-info in between these import groups?

src/text_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+// ---------------------------------------
+// TESTS FOR SPLITTING TEXT
+// ---------------------------------------
+
+test("Test: splitting a single sentences into List of words: ", () => {
```
Khazuar: It's clear these are tests, no need to name them "Test: ...".

src/text_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+	const result = sentenceIntoWordList("Hello, this is a test");
+	const expectedResult = ["Hello", "this", "is", "a", "test"];
+	expect(result).toEqual(expectedResult);
+});
+
+
+
+// ---------------------------------------
+// TESTS FOR EXPRESSIONS-LOGIK
```
Khazuar: LOGI**C**

src/text_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import ReactDOM from "react-dom";
```
Khazuar: I don't think you need these imports, nor does the file need to be a `tsx` file. Make it a `ts` file instead.

src/text_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+	const expectedResult = false;
+	expect(result).toEqual(expectedResult);
+});
+
+test("Test: if amount of left and right brackets are different.", () => {
+	const bracketList = ["[", "[", "[", "[", "[", "]", "]"];
+	const result = checkAmountOfBrackets(bracketList);
+	const expectedResult = true;
+	expect(result).toEqual(expectedResult);
+});
```
Khazuar: Consider using the describe-it-style:
```
describe("checkAmountOfBrackets", () => {
  it("detects wrong amount of brackets", () => {
    expect(checkAmountOfBrackets("..."))
      .toEqual(false);
  });
  it("", () => {
     expect ...
  });
});
```

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,140 @@
...
+// ---------format the input of a user into a full list of senteces including a list of words, formatted expressions, and paragraph marker-------
+export function textToListFormat(input: String) {
+
+	input = input.replace("$", "$EXPRESSIONMARKER");                            // prepare expressionsdetector; this is needed as marker to finde the expressions after the $ disappear by calling the split-method
+	input = input.replace("\n", "\n PARAGRAPHMARKER.");                         // prepare paragraphmarker
+
+	const separator = /[\n.!?$]+/g;                                              // List all the values the input will get splitted; \n will split paragraphs, $ will split expressions, the rest split sentences
+
+	const splittedText = input.split(separator);
+	const temp = [];
```
Khazuar: This is typed as `any[]`. Please add an explicit type.

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,140 @@
...
+				// temp.push("Sentence_" + (sentenceCount + 1));                // |optional
+				// sentenceCount++;                                             // |optional
+				const ListOfWords = sentenceIntoWordList(element);
+				temp.push(ListOfWords);
+			}
+
+		}
+
+	}
+	const listFormat = temp;
```
Khazuar: why?

src/text_formatter/text_formatter.ts
Khazuar: and a better name.

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,140 @@
...
+
+// --------------------------------------
+// @author: Ronja K.
+// ---------------------------------------
+
+// ---------format the input of a user into a full list of senteces including a list of words, formatted expressions, and paragraph marker-------
+export function textToListFormat(input: String) {
+
+	input = input.replace("$", "$EXPRESSIONMARKER");                            // prepare expressionsdetector; this is needed as marker to finde the expressions after the $ disappear by calling the split-method
+	input = input.replace("\n", "\n PARAGRAPHMARKER.");                         // prepare paragraphmarker
```
Khazuar: What if someone happens to write `PARAGAPHMARKER` into their input?

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,140 @@
...
+
+	const separator = /[\n.!?$]+/g;                                              // List all the values the input will get splitted; \n will split paragraphs, $ will split expressions, the rest split sentences
+
+	const splittedText = input.split(separator);
+	const temp = [];
+	// let sentenceCount = 0;                                                  //  |optional
+	// let expressionCount = 0;                                                //  |optional
+	// let paragraphCount = 0;                                                 //  |optional
+	for (let index = 0; index < splittedText.length; index++) {
```
Khazuar: `for (const text of splittedText) { ...`

src/text_formatter/text_formatter.ts
Khazuar: You can just split multiple times:
```
const result: any[] = [];
for (const paragraph of input.split("\n")) {
  const sentences: any[] = [];
  for (const sentence of paragraph.split(...) {
    ...
  }
  result.push(...sentences, [ "abs" ]);
}
```

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,140 @@
...
+
+	}
+	const listFormat = temp;
+	return listFormat;
+}
+
+// ---------format the words from a sentence-string into a list of words-------
+export function sentenceIntoWordList(input: String) {
+	const separator = /[ ,$]+/; // List all the values where the content of a sentence will get splitted: whitespace[ ] and comma[,]
+	const splittedSentenceIntoListOfWords = input.split(separator);
```
Khazuar: Instead of splitting, you could just scan for words: https://regex101.com/r/w47nEF/1

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,158 @@
...
+* ---------------------------------------
+* @param a input string of an expression where all operator has been formatted
+* @return a string array where each element is a element of the expression
+*/
+export function expressionFormatter(input: string) {
+	const detectedExpression = detectExpressionElements(input);
+	const separators = /(bracketLeft|bracketRight|equivalence|implication|negation|conjunction|disjunction)+/g;
+	const splittedExpression = detectedExpression.split(separators);
+	const temp = [];
+	for (let index = 0; index < splittedExpression.length; index++) {
```
Nifdee: You can also simplify this whole thing to something like 
```
return splittedExpressions
  .map(removeWhiteSpaces)
  .filter(x => x !== "")
  .map(replaceInputCharactersToReadablePrologCharacter);
```

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+* replace expression elements into readable prolog Code
+* ---------------------------------------
+* @param 
+* @return
+*/
+export function replaceInputCaractersToReadablePrologCharacter(input: string) {
+	input = input.replace(/(bracketLeft)/g, "\[");
```
Nifdee: These calls can also be chained:
```
return input
  .replace(/(bracketLeft)/g, "\[")
  .replace(...)
  .replace(...)
  ...;
```

src/text_formatter/text_formatter.ts
```diff
@@ -0,0 +1,158 @@
...
+	return input;
+}
+
+/**
+* check the expression if the amount of opened and closed brackets is correct; it can detect user mistakes; RETURN true if the user made a mistake;
+* ---------------------------------------
+* @param
+* @return
+*/
+export function checkAmountOfBrackets(formattedExpression: string[]) {
```
Nifdee: But what if the user does something like `)test(`? Shouldn't that also complain about incorrect brackets? I've actually seen the problem of detecting correct bracketing being used as a developer interview question :D (it's not hard to implement)

src/text_formatter/allowed_Expression_Detectors.ts
```diff
@@ -1,105 +1,39 @@
...
+	return formattedInputExpression
+		.replace(/(\[|\()/g, " bracketLeft ")
+		.replace(/(\]|\))/g, " bracketRight ")
+		.replace(/(<-->|<==>|<=>|<->)/g, " equivalence ")
+		.replace(/(->|-->|=>|==>)/g, " implication ")
+		.replace(/(neq|not|nicht|¬|-)/ig, " negation ")
+		.replace(/(and|und|&|∧)/ig, " conjunction ")
+		.replace(/(or|oder|\||∨)/ig, " disjunction ")
+		.trim()								// delete white spaces at the beginn and the end of an expression
+		.replace(/\s{1,}/g, ","); 			// each element of an expression should seperated with a comma
```
Khazuar: Instead of a string `"A,B,C..."` I'd rather return an array `["A","B","C"...]`:

- It's more convenient to work with ("how many tokens are there?", "get a subset of tokens", "Iterate over the tokens"...).
- It's more memory-efficient. Most string-implementations will require one consecutive piece of memory. Lists usually don't. Also, every operation on a string produces a copy of said string. 
- `string[]` or `Token[]` the more expressive return type (compared to `string` with some constraints on it's structure).

src/text_formatter/text_formatter.test.tsx
```diff
@@ -1,23 +1,16 @@
...
 });
 
-test("Test if the elements of a expression is detected correctly:", () => {
-	const result = detectExpressionElements("[AUNDTest[B<==>C]->DODERNOTNOTE]");
-	const expectedResult = " bracketLeft A conjunction Test bracketLeft B equivalence C bracketRight  implication D disjunction  negation  negation E bracketRight ";
+test("Test if the elements of a expression is detected and formatted correctly:", () => {
+	const result = formatExpressionElements("[AUNDTest[B<==>C]->DODERNOTNOTE]");
+	const expectedResult = "bracketLeft,A,conjunction,Test,bracketLeft,B,equivalence,C,bracketRight,implication,D,disjunction,negation,negation,E,bracketRight";
 	expect(result).toEqual(expectedResult);
 });
```
Khazuar: More tests please that cover corner cases. Empty input, excessive whitespace, incorrect input. If an error is supposed to happen, assert that as well!

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
  * @return a string-array where each element is one word
  */
 export function sentenceIntoWordList(input: String) {
-	const separator = /[ ,]+/; // List all the values where the content of a sentence will get splitted: whitespace[ ] and comma[,]
-	const splittedSentenceIntoListOfWords = input.split(separator);
-	const temp = [];
+
+	const splittedSentenceIntoListOfWords = input.split(wordSeparator);
+	const listOfWords = [];
```
Khazuar: Please add an explicit type.

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
...
 	for (let index = 0; index < splittedSentenceIntoListOfWords.length; index++) {
-		// remove all elements(senteces) which are a empty String after splitting
-		if (splittedSentenceIntoListOfWords[index].trim() == "") {
+		if (splittedSentenceIntoListOfWords[index].trim() == "") {	// remove all elements(senteces) which are a empty String after splitting
 			splittedSentenceIntoListOfWords.splice(index, 1);
 		} else {
-			temp.push(splittedSentenceIntoListOfWords[index].trim());
+			listOfWords.push(splittedSentenceIntoListOfWords[index].trim());
```
Khazuar: Put `splittedSentenceIntoListOfWords[index].trim()` into a variable.

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
  * @return a string-array where each element is one word
  */
 export function sentenceIntoWordList(input: String) {
-	const separator = /[ ,]+/; // List all the values where the content of a sentence will get splitted: whitespace[ ] and comma[,]
-	const splittedSentenceIntoListOfWords = input.split(separator);
-	const temp = [];
+
+	const splittedSentenceIntoListOfWords = input.split(wordSeparator);
+	const listOfWords = [];
 	for (let index = 0; index < splittedSentenceIntoListOfWords.length; index++) {
```
Khazuar:
```typescript
for (const word of splittedSentenceIntoListOfWords) { 
  ...
}
```

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
...
-		// remove all elements(senteces) which are a empty String after splitting
-		if (splittedSentenceIntoListOfWords[index].trim() == "") {
+		if (splittedSentenceIntoListOfWords[index].trim() == "") {	// remove all elements(senteces) which are a empty String after splitting
 			splittedSentenceIntoListOfWords.splice(index, 1);
```
Khazuar: Why do you remove these "bad" elements from `splittedSentenceIntoListOfWords`? You're putting the "good" ones into `listOfWords` already. And you don't need `splittedSentenceIntoListOfWords` anymore.

Khazuar: And btw, repeated splicing is just inefficient, since small lists in JS will be ArrayLists.

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
...
 		} else {
-			temp.push(splittedSentenceIntoListOfWords[index].trim());
+			listOfWords.push(splittedSentenceIntoListOfWords[index].trim());
 		}
 	}
```
Khazuar: This whole loop could be: `const listOfWords = splittedSentenceIntoListOfWords.map(w => w.trim()).filter(w => w.length > 0);`. Damn easy to read. Relatively efficient as well.

src/text_formatter/text_formatter.ts
```diff
@@ -57,42 +47,39 @@ export function textToListFormat(input: String) {
...
 * @return a string array where each element is a element of the expression
 */
 export function expressionFormatter(input: string) {
-	const detectedExpression = detectExpressionElements(input);
-	const separators = /(bracketLeft|bracketRight|equivalence|implication|negation|conjunction|disjunction)+/g;
-	const splittedExpression = detectedExpression.split(separators);
-	const temp = [];
+	const detectedExpression = formatExpressionElements(input);
+
+	const splittedExpression = detectedExpression.split(expressionSeparator);
```
Khazuar: You wouldn't need to split here if you returned an array from `formatExpressionElements`.

Khazuar:
I think you should separate your error pipeline work out into a new branch starting from master. I'd be glad to help you with that.

PR #25, opened 2019-05-17
RonjaKaehne/textFormatter

RonjaKaehne:


RonjaKaehne:
still got a Problem with the prolog.test.ts build

Korosensei42:
Sorry I just now get to review, but I don't think a commitmessage is the right place to ask for this. For my part: I get quicker to such things, if you message in Slack.

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
+import { addIssue } from "../issueHandling/issueMapping";
+const allowedExpressionToken = /(bracketLeft|bracketRight|equivalence|implicationRight|implicationLeft|negation|conjunction|disjunction|equal|addition|subtraction|division|multiplication)+/;
+const logicConnector = /(conjunction|disjunction|equivalence|implicationRight|implicationLeft)+/;
+const nonLogicToken = /(bracketLeft|bracketRight|negation)+/;
```
Korosensei42: Why is negation considered a nonLogicToken? I suggest you add negation to your logicConnector, even though it's a unary prediacte. Then a more suitable name for nonLogicToken would maybe be brackets

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+	* ---------CurrentExpressionFormatter------------
+	* @param expression written by a user
+	* @return a formatted expression as String
+	*/
+export function expressionFormatter(expression: string) {
+	const preFormattedExpression: string[] = preFormatExpressionFromImput(expression);
+	errorDetector(preFormattedExpression);
+	// TO DO SOME STUFF WITH THE TOKENS AND THE ARRAY; FORMAT THE ARRAY AND CATCH ERRORS
```
Korosensei42: Just curious: Why do you use capslock here? Does it fulfil some purpose?

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+	const finalFormattedExpression: string[] =
+		replaceExpressionElementsIntoPrologCode(preFormattedExpression);
+	return finalFormattedExpression;
+}
+
+/**
+ * 
+ * @param expression as a input string typed by user
```
Korosensei42: Even though @param gets generated automatically for the comment, there is no need for it, since one
can just read the method body to see the parameters.

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+		.replace(/\s{1,}/g, ",");
+	const splittedExpression: string[] = formattedInputExpression.split(/[,]+/g);
+	return splittedExpression;
+
+}
+
+/**
+	* @param formattedExpression
+	* @return finalExpression where exrpession elements got replaced with readable prolog code elements
+	*/
```
Korosensei42: *'s are not alligned. But that may have to do with browserdisplaystyle of github.

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+export function errorDetector(preFormattedExpression: string[]) {
+	detectBracketErrors(preFormattedExpression);
+	detectMissingStatementsOrConnector(preFormattedExpression);
+}
+
+/**
+	* check the expression if the amount of opened and closed brackets is correct; it can detect user mistakes;
+	* @param
```
Korosensei42: Nit: Check the expression for an equal amount of opened and closed brackets.

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,147 @@
...
+	* @return
```
Korosensei42: You could just delete this empty return.

src/input_formatter/text_formatter.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+	replaceExpressionElementsIntoPrologCode,
+} from "./expression_formatter";
+import { sentenceIntoWordList, textFormatter } from "./text_formatter";
+
+test("splitting a single sentences into List of words: ", () => {
+	emptyIssueList();
+	const result = sentenceIntoWordList("Hello, this is a test");
+	const expectedResult = ["Hello", "this", "is", "a", "test"];
+	expect(result).toEqual(expectedResult);
+});
```
Korosensei42: In the future you could use the describe-style for your tests. It makes what's tested more intuitive and your tests get more readable. This can for example look like this (taken from detectWrongSyntax.test.ts):

```
describe("logSingleWord", () => {
	it("Logs a ordinary word correctly", () => {
		const position: Position = {fromIndex: 0, toIndex: 4};
		const issues = logSingleWord("Hello", position);
		expect(issues).toEqual({
			code: issueJson.INVALID_WORD.message,
			message: `${"Hello"} von Stelle ${position.fromIndex} bis ${position.toIndex} ist ein unerlaubtes Wort! \n`,
			position: {
				fromIndex: position.fromIndex,
				toIndex: position.toIndex,
			},
			severity: issueJson.INVALID_WORD.severity,
		});
	});
```

You can even use multiple describes. Here's an example from diprocheInterface.test.ts:

```
describe("Error cases", () => {
		describe("Predicate is invalid", () => {
			it("Adds none of the regular predicates", () => {
				const mode = "Just a random text" as Mode;
				expect(addPredicate("", mode)).toEqual("Es gibt kein passendes Prädikat.");
			});
		});
	});
```

That is especially useful to distinguish different kinds of testcases.

Korosensei42:
Could you please add a test for the case someone writes "randomText not"? So, a text where someone uses the negation-connective, but doesn't negate something afterwards?

src/input_formatter/text_formatter.ts
```diff
@@ -0,0 +1,45 @@
+import { expressionFormatter } from "./expression_formatter";
+
+const inputSeparator = /([.!?]|\n|\$[^$]*\$)+/g;
+const wordSeparator = /[ ,]+/;
+
+/**
+ * this functions is the main function to format the full input string into a javastring array including formatted exrpessions
```
Korosensei42: What is the difference between a javastring array and a string array? Sorry, if this is common knowledge.

Korosensei42:
I do understand your code and just added some nitpicky comments. Otherwise semaphor says, that you have some linting errors. If you fix those, I don't have anything else to add (at least I don't know of anything to add right now)

src/input_formatter/expression_formatter.ts
Khazuar: Well, if the parameter-documentation includes a helpful description of what the parameter is for, it's legitimate.

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,155 @@
...
+	let index = 0;
+	while (index < preFormattedExpression.length) {
+		if (preFormattedExpression[index].match(allowedExpressionToken)) {
+			const element: string = replaceASingleExpressionElementIntoPrologCode(preFormattedExpression[index]);
+			finalFormattedExpression.push(element);
+		} else {
+			finalFormattedExpression.push(preFormattedExpression[index]);
+		}
+		index++;
+	}
```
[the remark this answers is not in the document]
Khazuar: Even better:
```typescript
foreach (const (index, expression) in preFormattedExpression.entries()) {
...
}
```

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,155 @@
...
+		} else {
+			if (foundStatement === true) {
+				addIssue("MISSING_CONNECTOR");
+			}
+			foundConnector = false;
+			foundStatement = true;
+			foundNegation = false;
+			index++;
+		}
+	}
```
Khazuar: for- or foreach-loop.

RonjaKaehne: it had a reason why i used a while loop, because with a index i can count faster through one loop after i matched an if or else if case

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,155 @@
...
+	let index = 0;
+	while (index < formattedExpression.length && bracketCount >= 0) {
+
+		if (formattedExpression[index].match(/(\[|bracketLeft)/)) {
+			bracketCount++;
+		} else if (formattedExpression[index].match(/(\]|bracketRight)/)) {
+			bracketCount--;
+		}
+		index++;
+	}
```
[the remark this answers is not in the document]
RonjaKaehne: i dont know how to use a for loop here when i use a second condition. btw why i should try to avoid while loops? just because to safe some lines of code?

src/input_formatter/expression_formatter.ts
```diff
@@ -0,0 +1,155 @@
...
+const logicConnector = /(conjunction|disjunction|equivalence|implicationRight|implicationLeft)+/;
+const bracket = /(bracketLeft|bracketRight)+/;
+const negation = /(negation)+/;
+
+/**
+ * main method to format expressions
+ * @param expression written by user
+ * @return a formatted expression
+ */
+export function expressionFormatter(expression: string) {
```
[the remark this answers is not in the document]
Khazuar:
```typescript
export function expressionFormatter(expression: string): string[] {
```

src/input_formatter/text_formatter.ts
```diff
@@ -7,7 +7,8 @@ const wordSeparator = /[ ,]+/;
  * this functions is the main function to format the full input string
  * into a string array including formatted exrpessions
  * @param input from user
- * @return the formatted text inculding formatted expressions
+ * @return the formatted text as string[] inculding
+ * formatted expressions, list of words and paragraph marker as empty Array
  */
 export function textFormatter(input: string) {
```
Khazuar: export function textFormatter(input: string): string[] {

src/input_formatter/text_formatter.ts
```diff
@@ -28,9 +29,8 @@ export function textFormatter(input: string) {
 
 /**
  * format the words from a sentence-string into a list of words
- * ---------------------------------------
  * @param a string including words
- * @return a string-array where each element is one word
+ * @return a string[] where each element is one word
```
Khazuar:
```typescript
export function sentenceIntoWordList(input: string): string[] {
```

src/input_formatter/expression_formatter.ts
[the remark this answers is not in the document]
Khazuar: For-loops have multiple advantages over while-loops:
* they combine declaring the running-variable, initializing it, incrementing it and checking the loop-condition all in one line, in a form, that is known to all programmers. Makes it easier to read.
* they are super-descriptive about what they are doing. They say "I am looping starting from here until there". That makes it easier to a reader to understand what your code actually does.
* they make it harder to get stuck in an infinite loop. With while-loops, the *change* to the loop-condition is happening somewhere within the body. If that body has multiple branches, it's difficult so always make sure your condition actually gets changed and to maintain it's correctness.

For similar reasons, `foreach`-loops are superior to for-loops:
* they don't declare any more variables than they need to. They also hide the complexity of actually iterating your collection completely away.
* they are super-short and point out your intentions as clearly as possible. "I am iterating this collections elements". 
* It's about impossible to build a non-terminating foreach-loop, unless you build an infinite iterable (please don't). 

I don't differentiate between a foreach-loop and the foreach-function here, and I have no strong opinion on what to use. Please just be consistent, and don't be the first to insert a foreach-loop in a function which only contained foreach-function-calls until then.

src/input_formatter/expression_formatter.ts
Khazuar: But you do see that it's pretty easy to build a non-terminating loop here? Are you sure you covered all cases and this code will not loop indefinitly for some input? If you have to use a while-loop somewhere, it's your job to make it clear for readers that it cannot not stop.

RonjaKaehne: yes im sure because in each loop each if/ if else/else case has its own index increment line so its impossible not to reach the breakout condition.

src/input_formatter/expression_formatter.ts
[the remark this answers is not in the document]
RonjaKaehne: thx for ur advices. i finally changed this loop into the for each loop deleting the second break condition

src/input_formatter/expression_formatter.ts
Khazuar: where is the difference to
```typescript
foreach (const expr of formattedExpression) {
		if (expr.match(bracket)) {
			continue;
		} else if (expr.match(negation)) {
			foundNegation = true;
			continue;
		} else if (expr.match(logicConnector)) {
			if (foundNegation === true) {
				addIssue("MISSING_STATEMENT_AFTER_NEGATION");
			} else if (foundConnector === true) {
				addIssue("MISSING_STATEMENT_INSIDE");
			}
			foundConnector = true;
			foundStatement = false;
			foundNegation = false;
			continue;
		} else {
			if (foundStatement === true) {
				addIssue("MISSING_CONNECTOR");
			}
			foundConnector = false;
			foundStatement = true;
			foundNegation = false;
			continue;
		}
	}
```
The `continue`-statements are not even necessary, but this way the first case isn't left empty.

PR #27, opened 2019-05-19
Fix error

TimothyGillespie:
Just a one liner which leads to a wrong error message.


TimothyGillespie:
@Khazuar I hope I understood you correctly with these changes.

PR #28, opened 2019-05-19
Timothy gillespie/tau prolog import function

TimothyGillespie:
Please don't feel urged to review my code on a sunday. 

This will implement an import function for prolog files.
When importing a prolog file it will be converted into PrologSession.
Also when there is a ":-use_module(importedModule)." it will now actually import the user made module which wasn't supported beforehand.

This functionality is useful to integrate the diproche program and to add functionalities which are not integrated by tau-prolog.


src/util/prolog/prologSession.ts
```diff
@@ -1,12 +1,32 @@
...
+		constructor(program: string, defaultPath: string = __dirname, limit?: number) {
 				this.session = new type.Session(limit);
+
+				let matchGroups: string[] | null;
+				let importedCode: string;
+
+				while ((matchGroups = fetchPrologImportsRegExp.exec(program)) !== null) {
+					 importedCode = fs.readFileSync(defaultPath + path.sep + matchGroups[2] + ".pl", "utf-8");
+					 program = program.replace(fetchPrologImportsRegExp, importedCode);
+				}
```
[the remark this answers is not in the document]
Khazuar: I understand you implement the importing by inlining. But if A imports B and C, and both B and C import D, D is included twice in your result. You should consider adding a `Set<string>` of filenames to your algorithm that you already imported.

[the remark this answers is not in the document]
Khazuar: I guess the order in which stuff is eventually imported doesn't matter? The prolog-programs lines can be reordered however?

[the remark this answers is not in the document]
Khazuar: You don't need to use `exec` and `replace` both, but you can do instead something like this:

```typescript
var program = program.replace(fetchPrologImportsRegExp, match => {
    return fs.readFileSync(defaultPath + path.sep + match[2] + ".pl", "utf-8");
});
```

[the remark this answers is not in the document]
TimothyGillespie: The order of the imports almost doesn't matter. There are examples where an import in a different order would break the code or change the order of the answers. If runtime can be improved by not strictly keeping the import order it would probably make sense to do so.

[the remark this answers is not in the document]
Khazuar: If the order of imports *does* matter, you probably want to sort all the imports based on their partial order. The algorithm for that is simple: Determine the complete dependency graph of all imports (A imports B, B imports C, B imports C, ...), then start picking files (A, B, C ...) that have no dependencies in the graph (import other stuff). Then remove that node and all import-edges. Add the contents of the files to your result in the order you pick them from the graph. *Tada*  
This will solve the problems with duplicate imports and circular dependencies will be obvious to spot.

TimothyGillespie:
Is still missing the suggested, more sophisticated import functionality therefore not yet ready for review.

src/util/prolog/prolog.test.ts
```diff
@@ -122,3 +123,27 @@ describe("PrologResult.getResultArray()", () => {
 			expect(results).toStrictEqual([["grandmotherm"], ["grandmotherf"], ["false"]]);
 		});
 });
+
+describe("PrologSession.importFile()", () => {
+	test('Imports the family.pl File Correctly"', async () => {
```
Nifdee: It'd be nice to describe what case you're testing; so instead of just specifying the file you're trying to use, explain what behavior you're testing by using that file - in this case e.g. `can import a file that doesn't import any other modules`, in the next one `recursively imports the contents of modules used in a file`; not sure what you're testing in the third case but I guess you get the idea :D

src/util/prolog/prologSession.ts
```diff
@@ -1,13 +1,37 @@
+import fs from "fs";
+import path from "path";
 import { format_answer, type } from "tau-prolog/modules/core";
 import {PrologResult} from "./prologResult";
 
+// RegExp group one will be the file to be imported.
+const fetchPrologImportsRegExp = /:-( |)use_module\(([a-zA-Z1-9]*)\) *\./g;
+const fetchModuleDeclerationsRegExp = /:-( |)module\([^)]*\) *\./g;
+
+export function importFile(relativePath: string): PrologSession {
```
Nifdee: `relativePath` means relative to what? The file from which I'm calling the function, or this file, or the location from which the program is run, or...? That's information for which it'd be worth creating a comment.

TimothyGillespie:
The test on semaphoreci failed just now because of different EOL Conventions between UNIX and Windows (I assume semaphore uses UNIX). I will change the methods so they will fit both conventions.

src/util/helperFunctions.ts
```diff
@@ -0,0 +1,44 @@
...
+		vertexesWithNoInEdge.forEach((vertex: string) => {
+			vertexesWithNoInEdge.delete(vertex);
+			order.push(vertex);
+		 graph.forEach((edge: string[]) => {
+				if (edge[0] === vertex) { graph.delete(edge); }
+				vertexesWithNoInEdge.add(edge[1]);
+			});
+		});
+	}
+ if (graph.size !== 0 ) { throw new Error("Cyclic Import"); }
```
Khazuar: You could return that residual graph here to let the user know what the circle is.

src/util/helperFunctions.ts
```diff
@@ -0,0 +1,44 @@
...
+	let matchGroups: string[] | null;
+
+	while ((matchGroups = pattern.exec(basis)) !== null) {
+		groupContent.push(matchGroups[group]);
+	}
+
+	return groupContent;
+}
+
+export function kahnAlgorithm(graph: Set<string[]>): string[] {
```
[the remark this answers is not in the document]
Khazuar: This exported method is modifying its input. That might be confusing to other users. 
You can clearly communicating that you are not modifying the input, by changing the input type to `ReadonlySet` or `Iterable`, then creating a modifiable set within the function scope.

src/util/helperFunctions.ts
```diff
@@ -0,0 +1,44 @@
+export function regexGroupToArray(basis: string, pattern: RegExp, group: number): string[] {
```
Khazuar: Could go to separate file `RegExpUtils.ts`

src/util/prolog/prologSession.ts
```diff
@@ -1,13 +1,65 @@
...
+	currentFileName: string,
+	currentGraph?: Set<string[]>) {
+
+		if (currentGraph === undefined) {
+			currentGraph = new Set<string[]>();
+		}
+
+		regexGroupToArray(program, fetchPrologImportsRegExp, 1).forEach((fileName: string) => {
+			// Note: The ! in currentGraph! is there to tell this object will not be undefined (or null)
+			// See: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html (Non-null assertion operator)
```
Khazuar: The non-null-assertion is unnecessary if you create a local variable `const graph = currentGraph;` right after the `undefined`-check in line 40 and use that variable in here instead.

src/util/prolog/prologSession.ts
```diff
@@ -1,13 +1,65 @@
...
+			currentGraph!.add([currentFileName, fileName]);
+			const absolutePath = defaultPath + path.sep + fileName + ".pl";
```
Khazuar: Please see https://nodejs.org/api/path.html#path_path_resolve_paths

src/util/helperFunctions.ts
```diff
@@ -0,0 +1,44 @@
...
+	 incoming.add(edge[1]);
+	});
+
+	incoming.forEach((vertex: string) => {
+		outgoing.delete(vertex);
+	});
+
+	return outgoing;
+
+}
```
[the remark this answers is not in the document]
TimothyGillespie: Understood! Just to clarify: We use camelCases for file names and PascalCases for Class names, right?

src/util/helperFunctions.ts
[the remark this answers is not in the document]
TimothyGillespie: I plan to add JSDocs for later in a different pull-request. I have delayed doing this so far because it didn't show the behavior I expected. I will look into this soon though.

src/util/getTopologicalOrder.ts
```diff
@@ -0,0 +1,56 @@
...
+			edgesToRemove.forEach((edge: Edge) => graph.delete(edge));
+		});
+		// Externalized Deletion
+		vertexesToRemove.forEach((toRemove: string) => vertexesWithNoInEdge.delete(toRemove));
+		vertexesToRemove.clear();
+	}
+
+	// If there are cycles in the graph the residual graph will be returned (which contains the cycles)
+	// This should be regarded as an error case
+	if (graph.size !== 0 ) { return graph; }
```
Khazuar: Sorry, my fault, I wasn't clear enough there.
Please don't *return* the errorneous state, but include it in a thrown error(-message).

src/util/getTopologicalOrder.ts
```diff
@@ -0,0 +1,56 @@
...
+	const edgesToRemove: Set<Edge> = new Set();
+	const vertexesToRemove: Set<string> = new Set();
+
+ while (vertexesWithNoInEdge.size !== 0) {
+		vertexesWithNoInEdge.forEach((vertex: string) => {
+			vertexesToRemove.add(vertex);
+			order.push(vertex);
+			graph.forEach((edge: Edge) => {
+					if (edge.origin === vertex) { edgesToRemove.add(edge); }
+					vertexesWithNoInEdge.add(edge.target);
```
Khazuar: You are currently iterating over `vertexesWithNoInEdge` (line 19).

src/util/prolog/prologSession.ts
```diff
@@ -17,11 +18,18 @@ export function importFile(relativePath: string): PrologSession {
...
 function resolveImports(program: string, defaultPath: string, currentFileName: string = "main"): string {
-		const dependencyGraph: Set<string[]> = getDependencyGraph(program, defaultPath, currentFileName);
-		const importList: string[] = kahnAlgorithm(dependencyGraph);
+		const dependencyGraph: Set<Edge> = getDependencyGraph(program, defaultPath, currentFileName);
+		const importList: string[] | Set<Edge> = getTopologicalOrder(dependencyGraph);
+
+		// Fetching an error; see in getTopologicalOrder.ts
+		if (importList instanceof Set) {
+			console.log("Residual Graph: " + importList.toString);
+			throw new Error("Cyclic Import. See the console for more information.");
```
Khazuar: Please throw this error from within `getTopologicalOrder`. Getting the residual graph as part of an error message is good enough, I don't think any code could reasonably "handle" these cases.

src/util/getTopologicalOrder.ts
TimothyGillespie: I think in this case it's fine, because I am only adding values(?). The issue with Strings and maybe Sets seemed to have been, that you possible jumped a value.

TimothyGillespie: Otherwise I can of course externalize the addition too, I just wanted to keep readability as good as possible.

PR #30, opened 2019-05-27
Export issue-json-key type as IssueCode

Khazuar:
A short suggestion on how to make your framework easier to use.

### DIFF
--- a/src/vocabularyChecker/detectWrongSyntax.test.ts
+++ b/src/vocabularyChecker/detectWrongSyntax.test.ts
@@ -1,7 +1,6 @@
 import React from "react";
 import ReactDOM from "react-dom";
-import { Severity } from "../checking/issue";
-import IssueCode from "../checking/issueCodes";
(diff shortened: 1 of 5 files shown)


src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -63,7 +62,7 @@ describe("logSingleWord", () => {
 		const position: Position = { fromIndex: 0, toIndex: 4 };
 		const issues = logSingleWord("Hello", position);
 		expect(issues).toEqual({
-			code: IssueCode.VocabularyIssue,
+			code: InvalidWord,
```
Khazuar: No, that won't work that way. `InvalidWord` as you imported it is a `{ message: string, severity: string }` and not a `string` as should be expected!

Khazuar: @Korosensei42

Issue #14, opened 2019-05-09
Create the Pre-Processing to Prolog Pipeline 


TimothyGillespie:
All the different functionalities (error detection, issue handling, text to prolog readable list, prolog interpreter) need to be connected.to a coherent system of forwarded and partially processed data with respective returns and feedback.

Possible Idea:
Vocabulary Error Detection -> Syntax Error Detection and Text To Prolog Readable List (text formatter) -> Issue Handler -> Prolog Interpreter.

Korosensei42:
For this Issue I would like to sketch my idea of how the pipeline should work alongside an example first.

Input: Text from User as a String. Example: "This text has a wrongWord and it holds that 2 += 2 5" where "wrongWord" is a wrong word.
**1) Vocabulary Error Detection:**
Tell the User, that "wrongWord" is not allowed and where "wrongWord" is. This can for example happen by highlighting "wrongWord" in the input.

Optimally: Make a suggestion for a replacement. However, this is not possible yet (later task for improving the vocabChecker)

Right now: Just let the User try things out until he has an accepted word.

Output: A String that consists only of allowed words. Example: "This text has a rightWord and it holds that 2 += 2 5"

**2) Syntax Error Detection**
Tell the User, that "2 += 2 5" is syntactically incorrect.

Optimally: Make a suggestion for improvement. However, I don't know if that is possible right now. Might have to wait.

Right now: Most likely treat it the same way as in step 1).

Output: A String consisting of syntactically correct expressions. Example: "This text has a rightWord and it holds that 2 + 2 = 5". 

Now, for the case that the user instead adds a new wrong word, start the pipeline at Step 1) again. Now with this output.

**3) Text to Prolog Readable List**
Take the String from before and transform it into a list.

Example: "[This, text, has, a, rightWord, and, it, holds, that, =[+[2,2],5]]].".

Output: List mentioned above.

**4) Issue Handler**
Tell the User what is still wrong. Most likely the remaining errors will be on a semantic level.

Ideally: Propose a specific solution to the user.

Right now: Just tell the user where his mistake is.

Example: Tell the User that "2 + 2 = 5" in his original text is semantically wrong and tell him where in his text this statement is. Wait for him to change "5" to "4". 

Now, it is possible that the user adds new wrong vocabulary or new wrong syntax. For this: Start the whole pipeline with the new input again at Step 1).

**5) Prolog Interpreter**
[...]

PR #31, opened 2019-05-27
Kweirauch / Setting up the Pipeline

Korosensei42:
This changeset takes all functions that currently generate Issues and runs all of them. Also it provides the possibility to throw all Issuemessages as an error message.

Also now it contains the whole vocabulary that is allowed as of right now.


TimothyGillespie:
To 1:
It seems the ifError function just throws anything as long as it is a Truthy Value. So almost everything. It wouldn't throw for null or undefined and I think some other, few values. It doesn't even seem to check if it actually is an error.
I found this: https://github.com/nodejs/node/issues/3398 (doesn't seem to praise ifError a lot)
and this: https://nodejs.org/api/assert.html#assert_assert_iferror_value (here you can see it just tries to anything)
I am not well versed in typescript/javascript error handling, but it doesn't seem to be recommended to use ifError? An alternative could be try-catch-blocks which we know from java though I am not sure if this is good practice. 
https://basarat.gitbooks.io/typescript/docs/types/exceptions.html
@Khazuar please veto me if I am saying something stupid.


To 2:
Regarding the diproche prolog results you can specify what output you would want and I will implement it, so you can just call a function and receive the expected output. You imported PrologResults but you will not need this one. This is just to handle diproche later.

To 3:
I would say define and specify the output you would expect and we will implement that interface. Meanwhile the specified function can just return dummy-output for your references.

Korosensei42:
> An alternative could be try-catch-blocks which we know from java though I am not sure if this is good practice.

Judging from your sources this seems like a better alternative. However, I don't know it's a good practice regarding this, either.

> Regarding the diproche prolog results you can specify what output you would want and I will implement it, so you can just call a function and receive the expected output.

I think this is something we should all discuss in detail together, since I think the prolog results themselves can be quite different in nature. I.e. does it still throw some syntaxerrors we don't cover or, most importantly, which kind of errors do we have on a semantical level? For the time being, I think a simple String "The proof is correct" or "The proof is wrong" should suffice.

> You imported PrologResults but you will not need this one. This is just to handle diproche later.

Thanks for the info! I was not sure about the use of this.

> I would say define and specify the output you would expect and we will implement that interface. 

Most outputs I expect are just a simple String, as described in #14 


Also thank you for the new PR. I will check it out right away.

TimothyGillespie:
Please see pull request #32 for a proposed interface for diproche.

TimothyGillespie:
Sorry, I had totally forgotten to check the information you have provided in #14.

src/pipeline/bundleFunctionalities.ts
```diff
@@ -0,0 +1,54 @@
...
+ * "2 + 2 = 5" is semantically wrong
+ *
+ * function showWrongSemantic(text: string) {
+ *
+ * }
+ *
+ */
+
+export function bundleEverything(text: string) {
+	ifError(showVocabErrors(text)); {
```
Khazuar: The correct way in jest to test that something throws an error:
```typescript
expect(() => { 
  // your code
}).toThrow();
```
you don't need to open a block if you call just a single function: `expect(() => fubar()).toThrow();`. You can also negate your assertion: `expect(() => fubar()).not.toThrow();`. Even though it's type-correct, I'd **not** write `expect(fubar).toThrow();`. Javascript function-objects are weird, and Typescripts type-system for them is not helping much. My advise: Whenever passing around "function-pointers", do so with an explicit lambda-expression.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,6 @@
+// Please specify and describe the return type for this function.
+// Input should be a string I can just run in diproche
+// (also the corresponding predicate like diproche_fo(...).)
+export default function getErrors(diprocheInput: string) {
```
Khazuar: What output would we expect from this function at all? All it can do is find issues in the proof provided to it. IMHO it can just add these issues to the global issue-list itself and return `void`.

Korosensei42:
I can't figure out why my first test doesn't work. Any ideas? I expect an error to be thrown and VSCode does actually throw the error, but the test fails.

TimothyGillespie:
Is the test failing on your local machine too or only on semaphoreci?

TimothyGillespie:
I think I have it.

TimothyGillespie:
I thought the issue was that it had to be `throw new Error(...);` instead of just `throw Error(...)`, but this didn't fix it. I will see if I can find the reason.

TimothyGillespie:
I think I have it now though. The issue seems to be, that you throw an error before you check for it by calling  `getErrors(...)` which then calls `getVocabErrors(...)` which is called before that too.

Please note that you should test errors thrown by using an arrow function like `expect(() => getVocabErrors(userInput)).toThrow(Error);`. You can assign it to a variable too this way like
`const expected = getVocabErrors(userInput);` to then put it in later.

I am not 100% sure but it seems you will need to do this to avoid throwing the error before that. It seems if you throw the error in the body of the test it will fail the test. You need to throw it as an `expect(...)` parameter. The technique with the arrow function seems to delay the error throw until you actually call the function (which contains the arrow throw).

Please see #37 for the version which I created. It doesn't implement what you wanted to implement but could give a direction on what I mean or how to get it running again.

Korosensei42:
@TimothyGillespie Whether I throw a new Error or not doesn't seem to make a difference. The thing with the arrowfunction however, did solve the problem. Thanks!

src/util/diproche/diprocheInterface.test.ts
```diff
@@ -0,0 +1,54 @@
...
+	 const userInput = "Angenommen A ist falshc. Dann ist A => B wahr.";
+	 const expected = () => getVocabErrors(userInput);
+	 expect(expected).toThrow(Error);
+	});
+});
+
+describe("addPredicate", () => {
+	describe("Normal cases", () => {
+		describe("Adding the predicates works for a normal string", () => {
+			it("Adds the prop-logic predicate correctly", () => {
```
Khazuar: How about 
```Typescript
// Top of file
const PREDICATES = ["propositionalLogic", "firstOrderPredicateLogic", "test"];

...
describe(...
  foreach (const pred of PREDICATES) {
    it(....
    it(....
    ...
  }
```

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,87 @@
...
+ * it needs to be added. For easier maintainability in case
+ * of an expanding predicate list, the corresponding predicate is
+ * added with respect to the current @param mode. This can, for
+ * example correspond to the concept of "Spielwiese". The user should
+ * be able to just stay inside a mode and not have to add this mode
+ * himself. Hence, the mode needs to be read out from the browser.
+ */
+export function addPredicate(userInput: string, mode: string) {
+	for (const predicate of modes.predicates) {
+		if (predicate.valueOf() === mode.valueOf()) {
```
Khazuar: What do you need these `valueOf` for?

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,87 @@
...
+			return replaceModeWithPredicate(mode).concat("(").concat(userInput).concat(").");
+		}
+	}
+}
+
+export function replaceModeWithPredicate(mode: string) {
```
Khazuar:
```suggestion
export function getPrologPredicateForMode(mode: string) {
```

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,87 @@
...
+		return "diproche";
+	}
+	if (mode === "firstOrderPredicateLogic") {
+		return "diproche_fo";
+	}
+	if (mode === "test") {
+		return "teste";
+	}
+	return "InvalidPredicate";
+}
```
Khazuar: If your `mode` were a union type (or an enum, which provides union-functionality), you could use exhaustive checking, see https://www.typescriptlang.org/docs/handbook/advanced-types.html Exhaustiveness checking.

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -150,12 +151,12 @@ describe("getAllIssues", () => {
 				position: {
 					fromIndex: 0,
 					toIndex: 9,
-				},
-				severity: 3,
+					},
```
Khazuar:
```suggestion
				},
```

src/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -60,41 +60,43 @@ describe("getInvalidWords", () => {
...
+			severity: Severity.FatalError,
 		});
 	});
 
-	/**
-	 * The next two testcases should never actually appear in practice. Since an empty word and a single whitespace
-	 * are not classified as "words" for our purposes. For the sake of completeness those cases are
-	 * tested nonetheless
-	 */
+/**
```
Khazuar: two * usually indicate JSDoc-comments.

src/util/diproche/diprocheInterface.ts
Korosensei42: Do you mean that I could define it as string | undefined?

Khazuar: Actually, `"propositionalLogic" | "firstOrderPredicateLogic" | "test"` is a type, too (one that can only take the values `"propositionalLogic"`,`"firstOrderPredicateLogic"` or `"test"`). You can give it a shorter name, too:
```typescript
type Mode = "propositionalLogic" | "firstOrderPredicateLogic" | "test";

export function replaceModeWithPredicate(mode: Mode): string {
```

If you do so, you do not have to return `InvalidPredicate` at the end, since the compiler can determine this line is never reached. It is good practice to throw a `UnexpectedError` there though, so we know about totally unexpected inputs as soon as possible:

```Typescript
export class UnexpectedError: Error {
  constructor(unexpected: never) {
    super(`Unexpected value '${unexpected}'!`);
  }
}
```

`never` is the Typescript type that a variable gets when previous checks ruled out all possible values. This usually only occurs when someone tampered with the typesystem (by using raw Javascript, casting or `any`).

Korosensei42: Understood. But which superconstructor should get called? I thought, superconstructors need to have some superclass as reference.

src/util/diproche/diprocheInterface.ts
Korosensei42: This was a leftover from some earlier debugging.

src/util/diproche/diprocheInterface.ts
Korosensei42: I just pushed the code as far as I understood you. One test now doesn't work anymore (which is understandable) and I get an error because of super (which I don't know how to solve currently).

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,96 @@
+import IssueCode from "../../checking/issueCodes";
+import { getAllIssues } from "../../vocabularyChecker/detectWrongSyntax";
+import modes from "./predicateList.json";
+
+type Mode = "propositionalLogic" | "firstOrderPredicateLogic" | "test";
+
+export class UnexpectedError {
```
TimothyGillespie:
```suggestion
export class UnexpectedError extends Error {
```

I think this is what @Khazuar meant this. This fixes the Error and should do what we want.

Korosensei42: You are right. Thanks!

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,96 @@
...
+import { getAllIssues } from "../../vocabularyChecker/detectWrongSyntax";
+import modes from "./predicateList.json";
+
+type Mode = "propositionalLogic" | "firstOrderPredicateLogic" | "test";
+
+export class UnexpectedError extends Error{
+	constructor(unexpected: never) {
+		super(`Unexpected value '${unexpected}'!`);
+	}
+}
```
Khazuar: This can go in a separate file, since it's usable all over the codebase.

src/util/diproche/diprocheInterface.ts
Khazuar: The tip by @TimothyGillespie did it ;)

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,96 @@
...
+export function getPrologPredicateForMode(mode: Mode) {
+	if (mode === "propositionalLogic") {
+		return "diproche";
+	}
+	if (mode === "firstOrderPredicateLogic") {
+		return "diproche_fo";
+	}
+	if (mode === "test") {
+		return "teste";
+	}
```
Khazuar: You could `switch` here.

Korosensei42:
The main problem right now seems too be, that the functions from issueMapping all reference the global variable inside there. Most likely I will need to solve this by just running all the functions to grow one big list of issues ans only at the end will I read the whole list.

Korosensei42:
@RonjaKaehne Could you give me a reminder as to which of your functions collects all the Issues corresponding to textformatting? I would like to add it to the pipeline.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
+import Issue from "../../issueHandling/issue";
+import { concatOneIssueList } from "../../issueHandling/issueMapping";
+import { getAllIssues } from "../../vocabularyChecker/detectWrongSyntax";
+import { UnexpectedError } from "./Errors";
+import modes from "./predicateList.json";
+
+export type Mode = "propositionalLogic" | "firstOrderPredicateLogic" | "test";
```
TimothyGillespie: A type will definitly work here too, but I think an enum would do better. The autocomplete function is more usable then the one for this type implementation. You don't have do handle strings which will indicate that this is not just randomly changeable and you can use a secondary value for them (if undefined they will just be 0, 1, 2 etc.)

src/util/diproche/predicateList.json
```diff
@@ -0,0 +1,10 @@
+{
```
TimothyGillespie: I think this json would become unnecessary. That it needs to have the keys listed twice makes is very unconvient to edit and add to as well.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
```
TimothyGillespie: I think you are describing the function too much in detail and trying to justify it.
The JSDoc should:

- Describe what the function achieves
- Document what the functions parameters should be with @param
- Document what the return will be (if it is not null) with @return
- Document which Exceptions it might throw and when (if any) with @throws (this isn't usually used a lot it seems, but it seems useful to me)

This is very useful because most IDEs will tell you the information that you have entered when you using the function somewhere else. It can also automatically be compiled to an actual documentation and give documentation a more structured character.

Justifications for this functions should usually be in non-JSDoc comments.

Here is an example
https://jsdoc.app/howto-es2015-classes.html

And here some more details on how to write a few things:
https://devhints.io/jsdoc

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
...
+		default: {
+			throw(UnexpectedError);
+		}
+	}
+}
+
+// Assuming "wrongWord" is not an allowed word, then a
+// vocaberror is an error like "the word "wrongWord"
+// in the text "This text contains a wrongWord".
+export function getVocabErrors(userInput: string): Issue[] | void {
```
TimothyGillespie: When is the return type a Issue[] and when void?

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
...
+		case "test": {
+			return "teste";
+		}
+		default: {
+			throw(UnexpectedError);
+		}
+	}
+}
+
+// Assuming "wrongWord" is not an allowed word, then a
```
TimothyGillespie: Examples should be described in tests.

src/util/diproche/diprocheInterface.test.ts
```diff
@@ -0,0 +1,89 @@
...
+				expect(addPredicate("", pred)).toEqual("diproche_fo().");
+			});
+			it("Adds test-predicate to an empty string", () => {
+				const pred = "test";
+				expect(addPredicate("", pred)).toEqual("teste().");
+			});
+		});
+	});
+	describe("Error cases", () => {
+		describe("Predicate is invalid", () => {
```
TimothyGillespie: In my opinion you are nesting describes too much. This makes it harer to read and in this case you put two extra describes for one test. I think normal cases, errors cases and limit cases don't have to be explicitely mentioned and can be described in the test name like "Will throw a SomeErrorException when the predciate is invalid".

Korosensei42: I wouldn't go as far saying one style is superior to the other, but in my mind this cascade of describes results in more structured testcases in general. I can see why it is redundant from your point of view, because it just adds more text to read while providing no additional functionality or testcoverage. However, in the case I add more and more tests later (which most likely will happen), one might find the repeated partition useful for understanding the testcases better.

Mostly it's just a choice of design, but for the time being, I prefer the nested describes. But I can see where you're coming from.

src/util/diproche/diprocheInterface.ts
Korosensei42: That was actually something I was wondering, too. As in: I know that examples should serve as tests, but I never figured, that examples should **exclusively** be tests. Here I figured a example would result in a better understanding in the code without having to look at the tests.

src/util/diproche/diprocheInterface.ts
Korosensei42: That was a leftover from some debugging I did before a commit. Good thing you catched it. void doesn't serve any purpose here.

src/util/diproche/diprocheInterface.ts
Korosensei42: You're right. This justification would be better fitted in some kind of design document.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
...
+ * of an expanding predicate list, the corresponding predicate is
+ * added with respect to the current @param mode. This can, for
+ * example correspond to the concept of "Spielwiese". The user should
+ * be able to just stay inside a mode and not have to add this mode
+ * himself. Hence, the mode needs to be read out from the browser.
+ */
+export function addPredicate(userInput: string, mode: Mode) {
+	for (const predicate of modes.predicates) {
+		if (predicate === mode) {
+			return getPrologPredicateForMode(mode).concat("(").concat(userInput).concat(").");
```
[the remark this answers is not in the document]
Korosensei42: Do you mean it is less readable because concat needs parantheses itself? Otherwise I don't see how concat makes it less readable. In fact, the intention behind using concat was better readability, so one can immediately understand, that two strings are added.

Is your argument behind + being more readable, that there is less text and no additional parantheses which could be confusing?

Either way I think this is mostly a choice of preference, so I don't really care if I change it to + or leave it at concat.

src/util/diproche/diprocheInterface.ts
Korosensei42: I see your point and will try the enum.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return "Es gibt kein passendes Prädikat.";
```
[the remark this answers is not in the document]
Korosensei42: Agreed.

src/util/diproche/diprocheInterface.test.ts
Korosensei42: When using your enum-based Ansatz, this particular error case got redundant anyways, since the linter doesn't allow for an invalid Predicate anyways. However, the topic of this choice of design remains.

src/issueHandling/issueMapping.ts
```diff
@@ -33,6 +33,12 @@ export function addIssueToIssueList(issue: Issue) {
 	foundIssues.push(issue);
 }
 
+export function concatOneIssueList(issues: Issue[]) {
+	for (const issue of issues) {
+		addIssueToIssueList(issue);
+	}
+}
```
Khazuar: This method breaks the encapsulation of the issueMapping-module. Please remove.

src/util/diproche/diprocheInterface.ts
Khazuar: Since we're writing Typescript (instead of Javascript), I think we should omit the type-info in JSDoc-comments though. This can be expressed better with the actual parameter-types / method-return-type.

src/util/diproche/diprocheInterface.ts
Khazuar: Users of your code will usually look for examples on how to use your method by using "Find References" (or similar) in their IDE. This way, they will find your tests really quickly.
Ideally though, your function needs no description at all, because its name+parameter(-names+-types)+returntype already tell the whole story. A quick one-line-explanaition can clarify something that is easily misunderstood. Nobody will read a longer method-comment though, when they don't understand at this point what's happening, it's more likely they will take a look at the method-code directly.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+export function addPredicate(userInput: string, mode: Mode) {
+	return mode.concat("(").concat(userInput).concat(").");
+}
+
+// Assuming "wrongWord" is not an allowed word, then a
+// vocaberror is an error like "the word "wrongWord"
+// in the text "This text contains a wrongWord".
+export function getVocabErrors(userInput: string): void {
+	concatOneIssueList(getAllIssues(userInput));
```
Khazuar: You should use the `addIssue`-method here. Since that method takes the issue-code only though, you will need to change `getAllIssues` to not return issues, but add the issue-codes via `addIssue` directly.

src/util/diproche/diprocheInterface.ts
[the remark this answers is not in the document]
TimothyGillespie: I meant it is less readable because of the parentheses.
I think using + to concate strings will not make it less readable to other developers though since this is pretty usual to do in javascript.

The only case were + should definitly be chosen over concat is when you call will call it many times for example in a loop, since + is faster than concat.

See here:
http://jsben.ch/D4itG

If you wanna address my comment and keep concat you could do something like this:
```
return getPrologPredicateForMode(mode)
  .concat("(")
  .concat(userInput)
  .concat(").");
```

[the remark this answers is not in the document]
Korosensei42: Interestingly enough I get a 2% faster result for concat when following your link, meaning concat is faster by the slightest bit. After running the tests a second time it was even 5% faster and after the 4th time running it, + was 5% faster. The variance seems to be even stronger, since after doing more test runs one was 20% faster than the other and then they swapped roles again.

The set of words on the left didn't seem to change. So I guess the results don't speak in favor of any one of those methods. Also, even one was faster, I don't think that's really the important thing, since I only use these methods twice, whereas the benchmark uses round about 500 words, thus using both methods around 500 times.

 However, since none of those methods seems to be faster and I most likely have author-bias, I will change it to +.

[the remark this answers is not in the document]
Khazuar: Benchmarking JavaScript is terribly complicated; trust me, I interned in the V8-team for 3 months, and if there's anything I learned, it's that you've got to take into account way to many things. We tested and benchmarked on countless different platforms and with different optimization strategies, and there was rarely something that was "obviously better in every possible way".
Therefore, I would put emphasis on readability. Your code doesn't seem to be performance sensitive anyways, so that should really be a no-brainer. Simplest solution (with a very reasonable performance, potentially even faster than `concat` or `+` (for number of parameters known at compile time):
```typescript
`${getPrologPredicateForMode(mode)}(${userInput})`
```
Here the compiler and optimizer can (at compile time) see how many things get combined and can potentially choose the most efficient approach right away. It is both short and pretty readable, too.

src/util/diproche/diprocheInterface.ts
Korosensei42: I don't understand why it would be disadventageous to immediately append a whole Issue[] to another one. If I understand the reason behind your suggestion I would agree on what I have to change, though.

src/issueHandling/issueMapping.ts
Korosensei42: What do you mean by encapsulation? I am not sure, if I have the same understanding of this word as you. In my mind it refers to the ability to manipulate the used data structures. Also I only use a method that was used before, so isn't this first method addIssueToIssueList breaking the encapsulation as well?
Also, I only use this new method on an Issue[] that I don't want to manipulate anyways.

Khazuar: IMHO we shouldn't allow to add `Issue`-instances to the issue-list directly.

Korosensei42:
A lot of tests I had previously got outdated when using the new intended way to handle issues. Therefore a lot is still untested. Also: We currently don't have any meaning of detecting semantic errors, afaik. So from a functional point of view, I guess I am done. I only have to test everything now.

Khazuar:
@Korosensei42 You need to merge master in here again.

Korosensei42:
I don't see where there are still changes requested, since everything was resolved in my opinion. Therefore I can't seem to merge master. GitHub tells me, that merging is blocked.

TimothyGillespie:
You want to merge into master?
In that case you not only need to address requested changes but to also have an approving review.
The requested changes don't resolve automaticaly too. You need to re-request a review.
Afaik: my requested changes wouldn't block you but @Khazuar's would. 
See https://help.github.com/en/articles/about-pull-request-reviews:

If required reviews are enabled and a collaborator with write, admin, or owner access to the repository submits a review requesting changes, the pull request cannot be merged until the same collaborator submits another review approving the changes in the pull request.

Korosensei42:
I might have expressed myself wrong. I wanted to merge the master branch into mine, only this time I wanted to try and do this via the browser. But it seems like I've misread something. Within the pull request there is no option to do it this way. Only the way you assumed.

Nonetheless I still wondered why some requests were left as necessary. That is clear now, after your answer.

src/util/diproche/diprocheInterface.ts
```diff
@@ -0,0 +1,59 @@
...
+export function getErrors(diprocheInput: string) {
+	getErrorsBeforeDiproche(diprocheInput);
+	// getErrorsAfterDiproche(diprocheInput);
+}
+
+export default function displayErrors(userInput: string) {
+	getErrors(userInput);
+	if (listAllIssues().length > 0) {
+		for (const issue of listAllIssues()) {
+			throw new Error(issue.message);
```
Khazuar: No. Throwing will leave this context the first time.

Korosensei42:
This changeset takes all functions that currently generate Issues and runs all of them. Also it provides the possibility to throw all Issuemessages as an error message.

src/util/diproche/diprocheInterface.ts
Korosensei42: Changed this. Also provided a description further down.

PR #34, opened 2019-05-29
Add JSDocs for the prolog related code

TimothyGillespie:


TimothyGillespie:
Not sure why the tests fail. They work on my machine and I only added comments + two return type declarations.

TimothyGillespie:
Die CI versucht aus irgendwelchen Gründen etwas zu machen, was ich allerdings nicht in diesem Branch implementiert habe sondern in TimothyGillespie/diproche.

TimothyGillespie:
You can't really do anything wrong reviewing these changes since these are only JSDocs additions. One could just give feedback on the documentation style and if they understand what the functions/methods do from the documention.

src/util/getTopologicalOrder.ts
```diff
@@ -3,7 +3,11 @@ export interface Edge {
 	target: string;
 }
 
-// Uses the Khan Algorithm loosely
+/**
+	* Uses the Khan Algorithm
+	* @param {Set<Edge>} graph - A graph represented as set of edges
```
Korosensei42: In your usecase, is it possible, that you have Vertices that are isolated, i.e. that are not connected to other vertices?
If so, modelling a graph as a set of edges would not suffice

src/util/prolog/prologResult.ts
```diff
@@ -9,22 +9,32 @@ const booleanAnswerRegExp = /^(true|false)( ;|,|.)$/;
...
+		*/
+	public getRawResults(): string[] {
 		return this.rawResults;
 	}
 
 	/**
-		* Generates (if needed) and returns the results.
-		*
-		*
+		* Creates and returns curated results
```
Korosensei42: What do you mean by "curated results"? (Note: Maybe my english vocabulary isn't good enough)

src/util/prolog/prologResult.ts
```diff
@@ -42,6 +52,10 @@ export class PrologResult {
 		return results;
 	}
 
+	/**
+		* Alternative way to access the results via an array.
+		* @return {Array<Array<string>>} First Level: i-1-th answer Second Level: Variable in query's order
```
Korosensei42: Nit: Isn't "i-1-st" correct?

src/util/prolog/prologResult.ts
```diff
@@ -51,12 +65,22 @@ export class PrologResult {
 		return results;
 	}
 
+	/**
+		* Receives the filled in values for a single variable
```
Korosensei42: Nit: filled-in (I think)

src/util/prolog/prologResult.ts
```diff
@@ -67,12 +91,18 @@ export class PrologResult {
 		return fetchAllMatchesForAGroup(source, pattern, 1);
 	}
 
+	/**
+		* @return {Set<string>} All variable which are assigned to in the raw results
+		*/
```
Korosensei42: Personally, I like to align the stars below each other, but I'd say that's up to preference.

src/util/prolog/prologSession.ts
```diff
@@ -29,6 +40,14 @@ function resolveImports(program: string, defaultPath: string, currentFileName: s
 		return importedPrograms + program;
 }
 
+/**
+	* Creates the dependency graph for imports of the prolog files recursively
+	* @param {string} program - The program code to look for imports
+	* @param {string} defaultPath - The path in which to look for the file names
```
Korosensei42: Does this result in a static path? Designwise this could later be adapted to be a dynamic path.

src/util/prolog/prologSession.ts
```diff
@@ -52,38 +71,53 @@ function getDependencyGraph(
...
-						});
-				});
-		}
+	public readonly session: any;
+
+	/**
+		* @param {string} program - The program code to run
+		* @param {string} defaultPath - Only needed for forwarding methode internally
+		* @param {string} fileName - Only needed for forwarding methode internally
+		* @param {number} limit - The maximum amount of resolving steps to do before aborting a query
```
Korosensei42: Maybe "maxIterations" would be a mote intuitive name.

src/util/prolog/prologSession.ts
Korosensei42: Edit: Didn't see the other variable relativePath earlier

src/util/getTopologicalOrder.ts
TimothyGillespie: The usecase is to import Files in the correct order so there will not be an isolated edge and if there were an isolated edge it wouldn't matter. Though I should add this to the description as I did with `getVertexesWithNoInEdge()` I should also correct the naming. There is no such thing as "Vertixes" oops.

I think I will call the nodes to make the names shorter.

src/util/prolog/prologResult.ts
TimothyGillespie: I guess it is not a frequently used word? It means the results are not just anything raw but they now more selective and prettied up.

Usually you curate an exhibition.

src/util/prolog/prologResult.ts
TimothyGillespie: It would be correct. I meant the i-th answer though. So I accidentally added the 1 it seems.

src/util/prolog/prologResult.ts
TimothyGillespie: Correct!

src/util/prolog/prologResult.ts
TimothyGillespie: Sometimes github changes the formatting completely for me. I don't know why.
I hope it looks proper when cloned?
I think you could call these kind of alignments an "act of violence". ^^

This is what they rather look like:
uplaod

src/util/prolog/prologSession.ts
TimothyGillespie: Would you think this is clear then or should the naming be changed still?

src/util/prolog/prologSession.ts
TimothyGillespie: That's true. I will change it to `maxSteps` since it will be shorter and these refer to the resolving steps as in the debugger if I am not mistaken.

TimothyGillespie:
@Korosensei42 Thanks a lot for reviewing!

src/util/prolog/prologSession.ts
Korosensei42: staticPath might also be viable, but defaultPath is just as understandable, So you can leave it be.

src/util/prolog/prologResult.ts
Korosensei42: Ah, so in my lingo I would declare them as "filtered" or something like that. But I guess your description fits at least as well.

src/util/getTopologicalOrder.ts
Korosensei42: If you would want to keep it close to the formal notion of a graph, you could also go with Vertices, which would be the correct plural iirc.

src/util/prolog/prologResult.ts
```diff
@@ -54,7 +54,7 @@ export class PrologResult {
 
 	/**
 		* Alternative way to access the results via an array.
-		* @return {Array<Array<string>>} First Level: i-1-th answer Second Level: Variable in query's order
+		* @return {Array<Array<string>>} First Level: i-th answer set Second Level: Variable in query's order
```
Korosensei42: Nit:
What do you mean by i-th answer **set**?
Do you really mean the verb "set"? If so: When is the answer set or by whom?

Or do you mean the noun "set", referring to the mathematical notion of a collection of elements? If so: Does your datatype `Array<Array<string>>` really have the properties of a set?

Korosensei42:
@TimothyGillespie You're welcome. Other than the few comments I just wrote, everything looks good to me. Except for the fact, that I don't know what Khan's algorithm does, but I can also just look it up later.
Maybe you could include some hyperlink explaining it as a comment for future developers?

src/util/prolog/prologResult.ts
Khazuar: It is probably the different tab-settings on the different systems. In our project, I set the tab-width to 2, whereas github uses 4 by default. You can add `?ts=2` to the end of the browser-location (like https://github.com/diproche/webinterface/pull/34/files?ts=2) to temporarily get the tabs "right".

src/util/getTopologicalOrder.ts
```diff
@@ -3,32 +3,36 @@ export interface Edge {
 	target: string;
 }
 
-// Uses the Khan Algorithm loosely
+/**
+	* Uses the Khan Algorithm (this implementation cannot work with isolated nodes)
+	* @param {Set<Edge>} graph - A graph represented as set of edges
+	* @return {Array<string>} The topologial oder of the graph
```
Khazuar:
```suggestion
	* @return {Array<string>} A topological ordering of the graph
```
There could be multiple such orderings after all.

src/util/getTopologicalOrder.ts
```diff
@@ -3,32 +3,36 @@ export interface Edge {
...
 			edgesToRemove.forEach((edge: Edge) => graph.delete(edge));
 		});
 		// Externalized Deletion
-		vertexesToRemove.forEach((toRemove: string) => vertexesWithNoInEdge.delete(toRemove));
-		vertexesToRemove.clear();
+		nodesToRemove.forEach((toRemove: string) => nodesWithNoInEdge.delete(toRemove));
+		nodesToRemove.clear();
 	}
 
 	// If there are cycles in the graph the residual graph will be returned (which contains the cycles)
```
Khazuar: This comment is redundant. Also, `graph.toString` is a function, not a `string`. You need to invoke it (add `()`).

src/util/prolog/prologResult.ts
```diff
@@ -9,22 +9,32 @@ const booleanAnswerRegExp = /^(true|false)( ;|,|.)$/;
...
 	constructor(rawResults: string[]) {
 		this.rawResults = rawResults;
 	}
 
-	public getRawResults() {
+	/**
+		* rawResults getter-Methode
+		* @return {Array<string>} The raw results
+		*/
+	public getRawResults(): string[] {
```
Khazuar: Either make the return type readonly (either by writing `readonly string[]` or `ReadonlyArray<string>`) or return a copy.

Khazuar:
@TimothyGillespie I think we should not include the parameter- and return-types in the JSDocs, since adding a type-annotation in Typescript is much better.

Khazuar:
Any blockers here? We should merge this soon.

src/util/prolog/prologResult.ts
TimothyGillespie: There is a mathematical set which follows more strict rules, but here I don't mean the mathematical definition. I wouldn't know a better description for this though I was thinking the same.

src/util/prolog/prologSession.ts
```diff
@@ -53,38 +72,53 @@ function getDependencyGraph(
...
+			return new PrologResult(rawResults);
+	}
+	/**
+		* Receives the answers for the query set by set
+		* @return The next result set
+		*/
+	private getNextAnswer(): Promise<string> {
+			return new Promise<string>((resolve) => {
+					this.session.answers((x: any) => {
+							resolve(format_answer(x));
```
Korosensei42: Nit: "x" is somewhat meaningless, but I think it's not really necessary to rename it, since the context describes it well enough already. But we should refrain from using such variablenames more often.

TimothyGillespie: Agreed

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,55 +1,94 @@
...
 }
 
-function removeDoubleWhitespaces(prologProgram: string) {
+/**
+	* To reduce code
+	* @param path - The program code to be curated
+	* @return Program code without double whitespace
+	*/
+function removeDoubleWhitespaces(prologProgram: string): string {
 	const fetchDoubleWhitespacesRegExp = /  */gm;
```
Korosensei42: This Regex seems to also detect single WhiteSpaces. Is this intended?

TimothyGillespie: You're right and that is not intended. Should have been a + instead and that is the case now. The last push made undid your approval though.

src/util/prolog/prologResult.ts
```diff
@@ -9,22 +9,32 @@ const booleanAnswerRegExp = /^(true|false)( ;|,|.)$/;
...
 const variableNameRegExp = /(^| )([A-Z]\w*)/g;
 
+/** Handles the results for a prolog query returned by PrologSession.executeQuery() */
 export class PrologResult {
 	private readonly rawResults: string[];
 	private results: Map<string, Array<string | boolean>> | undefined;
 
+	/**
+		* Simply forwards the raw results to its respective attribute
+		* @param {Array<string>} rawResults - The string array containing the results
```
Nifdee: To make documentation most useful, you don't need to describe the type or anything already contained in the name, but give additional information - e.g. what format those raw results should have, or where you would get them from (e.g. "The values received from ..." or "The prolog results in format ..."). All the documentation in this block I could have written without knowing what any of this code does; but I still don't know how to use this class / what to pass into the constructor.

src/util/prolog/prologResult.ts
```diff
@@ -9,22 +9,32 @@ const booleanAnswerRegExp = /^(true|false)( ;|,|.)$/;
...
 	constructor(rawResults: string[]) {
 		this.rawResults = rawResults;
 	}
 
-	public getRawResults() {
+	/**
+		* rawResults getter-Methode
+		* @return The raw results
+		*/
+	public getRawResults(): readonly string[] {
```
Nifdee: You could use `public readonly rawResults: ReadonlyArray<string>` as type instead of using an ordinary array. Then a) this getter method would be superfluous and b) you'd also enforce that no private method will ever modify these values either.

src/util/prolog/prologResult.ts
```diff
@@ -51,12 +65,22 @@ export class PrologResult {
...
+		* @return Fill in values in order of their return
+		*/
 	private getResultsFor(variable: string): string[] {
 		// Group 1 will be the content of the variable independent of the variable length.
 		const pattern = new RegExp(variable + variableAnswerPattern, "g");
 		return fetchAllMatchesForAGroup(this.rawResults.toString(), pattern, 1);
 	}
 
+	/**
+		* Curates and raw results so every variable has their own field and only their fill in values are given
```
Nifdee: Is there a word missing here? And what are fill in values?

src/util/prolog/prologResult.ts
```diff
@@ -51,12 +65,22 @@ export class PrologResult {
...
+		* @param variable - One line of a raw result
```
Nifdee: There is no param called `variable`...?

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,55 +1,94 @@
...
+	const fetchFileExtension = /(.+)\.[^\.]*/;
 	const fetch: string[] | null = fetchFileExtension.exec(path);
 	if (fetch === null) { return ""; }
 	return fetch[1];
 }
 
-export function removeModuleDeclarations(prologProgram: string) {
+/**
+	* Necessary because module declarations break tau-prolog
+	* @param path - The program code to be curated
```
Nifdee: *Räusper* if you copy-paste, then do it correctly! :D 
The parameter is called `prologProgram` in this function and the following ones...

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,55 +1,94 @@
 import {fetchAllMatchesForAGroup} from "../regExpUtils";
 import builtInPredicates from "./builtInPredicates";
 
-export function removeFileExtension(path: string) {
-	const fetchFileExtension = /(.*)\.[^\.]*/;
-
+/**
+	* @param path - The path to be curated
```
Nifdee:
```suggestion
	* @param path - The path of the prolog file to be curated
```

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,55 +1,94 @@
...
 }
 
-function removeEmptyLines(prologProgram: string) {
+/**
+	* To reduce code
+	* @param path - The program code to be curated
+	* @return Program code without empty lines and lines with only whitespace characters
+	*/
+function removeEmptyLines(prologProgram: string): string {
 	const fetchEmptyLinesRegExp = /\n(?:\s)*\n/gm;
```
Nifdee: It's slightly more performant and readable to declare all these regexes once at the top of the file instead of declaring them inside the function; it currently creates a new instance of them every time the function is called, which is unnecessary.

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,55 +1,94 @@
...
+	* To reduce code
+	* @param path - The program code to be curated
+	* @return Program code without parts that don't affect the behavior of the program
+	*/
+export function removeNonFunctionalities(prologProgram: string): string {
 	return removeComments(removeEmptyLines(removeDoubleWhitespaces(prologProgram)));
 }
 
+/**
+	* Renames predicates based on imports to avoid variable name fixVariableShadowingInImport
```
Nifdee: Should the last word really be `fixVariableShadowingInImport`?

src/util/prolog/prologStringManipulation.ts
```diff
@@ -62,6 +101,13 @@ export function fixVariableShadowingInImport(program: string, importedProgram: s
 	return identifyPredicatesUniquely(program, importedProgram, importedPredicates);
 }
 
+/**
+	* Renames the predicates in the imported file except for built-in and exported predicatesToRename
+	* @param program - The base program's code which is importing
+	* @param importedProgram - The imported program's code
+	* @importedPredicates - Predicates which are declared exported by the importedProgram
```
Nifdee:
```suggestion
	* @param importedPredicates - Predicates which are declared exported by the importedProgram
```

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: I see. They used to be all at the top of the file, but I decided to pull them into the function for readability and since these regular expressions couldn't be reused.

@Nifdee what would you think about declaring them outside of the function above the function signature like:

```
const fetchEmptyLinesRegExp = /\n(?:\s)*\n/gm;
function removeEmptyLines(prologProgram: string): string {
[...]
}
```

src/util/prolog/prologResult.ts
TimothyGillespie: Fill-in values are supposed to be what prolog returns as a possible variable value to make a predicate true. It could seem like an assignment, but strictly speaking it isn't so I tried maybe too hard to avoid that term.

TimothyGillespie:
@Nifdee Thank you for taking the time to give me valuable advice and hints. Very much appreciated! :)

src/util/regExpUtils.ts
```diff
@@ -1,3 +1,10 @@
+/**
+	* Fetches a RegExp's group's values by its number for all matches
+	* @param {string} textBasis - The text to be fetched from
```
Korosensei42: Why is the param here different to the one in the signature of the actual function?

PR #35, opened 2019-05-30
Integrate the diproche program

TimothyGillespie:


TimothyGillespie:
The previous errors were due to window specific cmd commands. I made them run cross-platform now by using the shx library which allows to use UNIX commands on all platforms.

https://github.com/shelljs/shx

.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+/src/util/diproche/diprocheProgram
```
Khazuar: I wouldn't put the diproche-program in the `src` folder if it's not actual source-code of this project.

src/util/diproche/diproche.ts
```diff
@@ -0,0 +1,16 @@
...
+	private readonly session: PrologSession;
+
+	constructor() {
+		this.session = importFile("./diprocheProgram/diproche.pl");
+	}
+
+	public async getFeedBack(input: string): PrologResult {
+		const results: PrologResult | Promise<PrologResult> =  this.session.executeQuery(input);
+		return results;
+	}
```
Khazuar: This is missing `Promise<..>` and `await`.

src/util/diproche/diproche.ts
```diff
@@ -0,0 +1,15 @@
...
+export default class Diproche {
+private readonly session: PrologSession;
+
+	constructor() {
+		this.session = importFile("../../.././readable_node_module_copies/diproche/diproche.pl");
+	}
+
+	public async getFeedback(input: string): Promise<ReadonlyMap<string, Array<string | boolean>>> {
+		const resultObject =  await this.session.executeQuery(input);
+		return new Promise(resolve => resolve(resultObject.getResults()));
```
Khazuar: This whole method should read:
```typescript
return await this.session.executeQuery(input);
```

TimothyGillespie:
Differences between SWI and tau-prolog forces us to pause this or find another solution since the current solution via a local SWI-Server is unsatisfactory.

PR #38, opened 2019-06-03
Remove unused imports

TimothyGillespie:
@Korosensei42 I think these are unusued imports and can therefore be removed or are they needed for something that I don't see right now?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,25 +1,27 @@
 {
-  "compilerOptions": {
-    "target": "es5",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
-    "allowJs": true,
-    "skipLibCheck": true,
-    "esModuleInterop": true,
-    "allowSyntheticDefaultImports": true,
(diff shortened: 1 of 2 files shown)


Korosensei42:
Indeed, they are unused right now. One could only expand the functionalities and/or add more tests, but right now no addition is necessary.

tsconfig.json
```diff
@@ -1,25 +1,27 @@
...
+		"strict": true,
+		"forceConsistentCasingInFileNames": true,
+		"module": "esnext",
+		"moduleResolution": "node",
+		"resolveJsonModule": true,
+		"isolatedModules": true,
+		"noEmit": true,
+		"jsx": "preserve",
+		"noUnusedLocals": true,
+		"noUnusedParameters": true
```
Khazuar: I added the last two lines.

Korosensei42:
Sorry I didn't comment on this earlier, but I oversaw that it is ready for review again. Overall it looks good to me. Got nothing to add right now.

PR #40, opened 2019-06-06
Timothy gillespie/prolog improve use module

TimothyGillespie:
I thought I had uploaded this branch too, but I didn't. My bad.

Tau-Prolog doesn't execute the diproche program yet and only returns "false" for the predicate, but doesn't return an error. Therefore I think there is an issue with overlapping predicate names. This branch is supposed to add identifiers to self-defined predicates to avoid an overlap of predicate names.


TimothyGillespie:
As discussed in the meeting I will be working on this to get it done asap.

TimothyGillespie:
The new function which will avoid shadow names of imports seems to work as intended. I will add more tests later. While integrating this functionality it seems I have broken the imports functionality. I will look into fixing this soon.

TimothyGillespie:
Found the error. I didn't add use_module and module to the list of built-in predicates. These have the special ":-" prefixed syntax which means they will break the code when executing it if the parameters or the predicate doesn't "make sense".
I hope this will solve the issues with the diproche integration now.

src/util/prolog/prolog.test.ts
```diff
@@ -177,4 +177,23 @@ describe("prologStringManipulation Module", () => {
...
+
+	test("fixVariableShadowingInImport() renames multiple predicates", async () => {
+		const results = PSM.fixVariableShadowingInImport("male(bob). male_1(karl).", "male(frank). male(hannes).");
+		expect(results).toEqual("male_2(frank). male_2(hannes).");
+	});
+
+	test("fixVariableShadowingInImport() work even with parameters containing braces", async () => {
+		const results = PSM.fixVariableShadowingInImport("male(b(ob)).", "male(ha(n(ne)s)).");
+		expect(results).toEqual("male_1(ha(n(ne)s)).");
+	});
```
Korosensei42: Could you give each function it's individual "describe" (as in using describe-cascades)? That makes it more readable, since you see, that some amount of tests all together test one function. Also within the tests for one function you could go for "describes" that group the tests by error-/edge-/normal cases.

src/util/prolog/prologSession.ts
```diff
@@ -4,7 +4,8 @@ import { format_answer, type } from "tau-prolog/modules/core";
 import getTopologicalOrder, {Edge} from "../getTopologicalOrder";
 import {fetchAllMatchesForAGroup} from "../regExpUtils";
 import { PrologResult } from "./prologResult";
-import { removeFileExtension, removeModuleDeclarations, removeNonFunctionalities } from "./prologStringManipulation";
+// PSM ~ prologStringManipulation
+import * as PSM from "./prologStringManipulation";
```
Korosensei42: I would prefer that you don't shorten a name that much. Imagine the file gets a lot larger and someone tests something, just to see that something in line 743 does not work. This person then proceeds trying to understand the code and reads "PSM". With this name that person then needs to jump around the file to look for that one comment you left.

src/util/prolog/prologSession.ts
```diff
@@ -23,7 +24,7 @@ function resolveImports(program: string, defaultPath: string, currentFileName: s
 		let importedPrograms: string = "";
 		importList.forEach((importFileName: string) => {
 			const currentImport = fs.readFileSync(path.resolve(defaultPath, importFileName + ".pl"), "utf-8");
-			importedPrograms = importedPrograms + currentImport;
+			importedPrograms = importedPrograms + PSM.fixVariableShadowingInImport(importedPrograms + program, currentImport);
```
Korosensei42: Just curious: How does the "+"-operator between importedPrograms and program work? Just a concatenation?

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
-const fetchModuleDeclaration = /:-( |)module\([^)]*\) *\.[\r\t\f\v ]*\n?/g;
-const fetchPartialLineCommentsRegExp = /(^[^\n%]+)%[^\n\r]*(\r?\n)/gm;
-const fetchFullLineCommentsRegExp = /^[\r\t\f\v ]*%[^\n]*\n/gm;
-const fetchEmptyLinesRegExp = /\n(?:\s)*\n/gm;
-const fetchDoubleWhitespacesRegExp = /  */gm;
+import {fetchAllMatchesForAGroup} from "../regExpUtils";
+import builtInPredicates from "./builtInPredicates";
 
 export function removeFileExtension(path: string) {
+	const fetchFileExtension = /(.*)\.[^\.]*/;
```
Korosensei42: We could consider being extra nice to future developers and comment every regex with a comment
to this website https://regexper.com/ containing the regex in question in the link (after an #). For example here:
https://regexper.com/#/(.*)\.[^\.]*/

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
 }
 
 function removeFullLineComments(prologProgram: string) {
+	const fetchFullLineCommentsRegExp = /^[\r\t\f\v ]*%[^\n]*\n/gm;
+
 	return prologProgram.replace(fetchFullLineCommentsRegExp, "");
 }
 
 function removePartialLineComments(prologProgram: string) {
+	const fetchPartialLineCommentsRegExp = /(^[^\n%]+)%[^\n\r]*(\r?\n)/gm;
```
Korosensei42: For me it is not totally clear what a partial line comment is. Could you give an example?

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
-const fetchPartialLineCommentsRegExp = /(^[^\n%]+)%[^\n\r]*(\r?\n)/gm;
-const fetchFullLineCommentsRegExp = /^[\r\t\f\v ]*%[^\n]*\n/gm;
-const fetchEmptyLinesRegExp = /\n(?:\s)*\n/gm;
-const fetchDoubleWhitespacesRegExp = /  */gm;
+import {fetchAllMatchesForAGroup} from "../regExpUtils";
+import builtInPredicates from "./builtInPredicates";
 
 export function removeFileExtension(path: string) {
+	const fetchFileExtension = /(.*)\.[^\.]*/;
+
```
Korosensei42: Nit: Is there a reason for this empty line? You seem to have a similar style with the other functions. While some people might consider the code more readable this way, some also consider it more "disconnected", as in: They feel like after the empty line something conceptually different should happen. But in reality it's just another command inside the same function. I consider myself to be one of the latter people, but mostly it's just a choice of design and I would also be okay if you leave it this way.

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
 	const fetch: string[] | null = fetchFileExtension.exec(path);
 	if (fetch === null) { return ""; }
 	return fetch[1];
```
Korosensei42: What happens if fetch is an array of size 1? Then the only entry it has is entry 0, isn't it? Returning entry 1 shouldn't work in that case.

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
 function removeFullLineComments(prologProgram: string) {
+	const fetchFullLineCommentsRegExp = /^[\r\t\f\v ]*%[^\n]*\n/gm;
+
 	return prologProgram.replace(fetchFullLineCommentsRegExp, "");
 }
 
 function removePartialLineComments(prologProgram: string) {
+	const fetchPartialLineCommentsRegExp = /(^[^\n%]+)%[^\n\r]*(\r?\n)/gm;
+
 	// group2 fetches the EOL standard in the file to fit both: Windows and UNIX
```
Korosensei42: I think it wouldn't hurt to type out the meaning of EOL. I, for my part, had to look it up. But it might just as well be common knowledge, which I didn't happen to have.

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
+	return identifyPredicatesUniquely(program, importedProgram, importedPredicates);
+}
+
+function identifyPredicatesUniquely(program: string, importedProgram: string, importedPredicates?: string[]): string {
+
+	// imports shouldn't be renamed
+	let predicatesToNotRename: string[];
+	// Checking for undefined based on this recommendation:
+	// https://basarat.gitbooks.io/typescript/docs/javascript/null-undefined.html
+	if (importedPredicates == null) {
```
Korosensei42: Just curious: Does this still work with `===` instead of `==`?

src/util/prolog/prologStringManipulation.ts
```diff
@@ -1,41 +1,122 @@
...
+			return toRename + "_" + identifier + parameters;
+
+		});
+	});
+
+	return importedProgram;
+}
+
+function getNonBuiltInPredicates(program: string): Set<string> {
+	// This doesn't take comments into account so in "%predicate()" would find predicate()
```
Korosensei42: Just curious: Is this how you want it to be in the end or is it your goal to change this later?

Korosensei42:
Overall your code looks fine to me. Just left you some comments about some sections I was not sure about and that maybe need some second thoughts or better explanations.

src/util/prolog/prologSession.ts
Khazuar: I think this is justified here though. Maybe `PSM` is not an ideal name, but I cannot think of anything that's better in every way.

src/util/prolog/prologSession.ts
Khazuar: I think string-"addition" is common enough to be understood.

src/util/prolog/prologStringManipulation.ts
Khazuar: Nah, if a developer wants to understand a regex further they'll do so. No need to add comments that will likely just desync at some point anyways.

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: With those I meant comments like this:
```
somefunction(x, y). // This is the partial line comment. It doesn't start at the beginning of the line.
```

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: My thinking here was that it does something different. The first part is just a setup which didn't even used to be inside the function but at the beginning of the file. The actually process happens only after.

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: Good question. I just checked the behavior of that.

It works out though. The variable `fetch` will always be either `null` or `string[1]` will be defined.
Index 0 refers to the whole match and index 1 refers to the first group. So one limit case would be as example:
`.pl` which will just evaluate to an empty string then. Group 1 will just be empty.
In case of no match it will return null.

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: I would have thought it is easier to find the term with EOL, but it isn't. I just tried googling it. I will include a link for more information on that since it isn't as easy to find as I thought it would be.

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: It doesn't. Normally we would get an error for using `==` instead of `===`, but if comparing to null it is allowed without me needing to change anything. So the conclusion is `null` and `undefined` have the same value, yet are they different objects.

src/util/prolog/prologStringManipulation.ts
TimothyGillespie: I didn't plan to change it. It was more of a note to justify why I call `removeNonFunctionalities(program)` in line 114.

The part with the new line error can be avoided with a different regexp, but it would be more vunerable to change it and less efficient. The current one is not very efficient already.

src/util/prolog/prolog.test.ts
TimothyGillespie: To group by functions totally make sense which I will implement. For different cases I will think about how to structur that. I should also split the tests over more files since they refere to different functionalities by now.

TimothyGillespie: I will revise the test cases later actually. I wanna work with these changes to make diproche run soon.

TimothyGillespie:
Opps. I didn't know I will make the approval go stale by adding something; makes sense though. My bad.

src/util/prolog/prologStringManipulation.ts
Khazuar: So change the regex to `/(.+)\.[^\.]*/` then.

src/util/prolog/prologStringManipulation.ts
Khazuar: Why don't you do `=== undefined` or `if(importedPredicates)` (checking the truthieness)?

PR #41, opened 2019-06-06
Add variable errorHere to line with error

Korosensei42:
The problem right now is, if you delete the "errorHere" variable from the Props-interface, you get a lintererror, since no empty interfaces are allowed. How should we handle this?

### DIFF
--- a/src/components/proofEditor/proofEditor.tsx
+++ b/src/components/proofEditor/proofEditor.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { checkProof } from "../../util/proofChecker";
 import styles from "./proofEditor.module.scss";
 
+// tslint:disable-next-line:no-empty-interface
 export interface Props {
 	errorHere: string;
 }
--- a/src/components/proofEditor/proofEditor.tsx
+++ b/src/components/proofEditor/proofEditor.tsx
@@ -3,7 +3,7 @@ import { checkProof } from "../../util/proofChecker";
 import styles from "./proofEditor.module.scss";
 
 export interface Props {
-
+	errorHere: string;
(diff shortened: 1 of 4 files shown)


src/components/proofEditor/proofEditor.tsx
```diff
@@ -3,7 +3,7 @@ import { checkProof } from "../../util/proofChecker";
 import styles from "./proofEditor.module.scss";
 
 export interface Props {
```
TimothyGillespie:
```suggestion
// tslint:disable-next-line:no-empty-interface
export interface Props {
```

This will ignore the empty interface linter error.

TimothyGillespie: Another option could be to remove the Interface entirely and change line 13 to `... extends React<State> {`

I think the interface is there to make the code more readable and to predefine State and Props and I think it helps. So I would keep it, since it let's us now that there are no props passed.

src/components/proofEditor/proofEditor.tsx
```diff
@@ -3,7 +3,7 @@ import { checkProof } from "../../util/proofChecker";
 import styles from "./proofEditor.module.scss";
 
 export interface Props {
-
+	errorHere: string;
```
TimothyGillespie:
```suggestion
```

This suggestion is just here so you can make the changes from GitHub and see if it works right away.

src/components/proofEditor/proofEditor.tsx
Korosensei42: I agree that disabling the linterrule for this would be one solution. I will apply it for now.

PR #42, opened 2019-06-11
Add parameter for issue message placeholder replacements

Khazuar:


Khazuar:
> Except for the question on Slack, I approve of this.
> For reference the question was:
> 
> How does `string|number` allow for some message like
> "Das Wort Schuh von Position 0 bis 5 ist unerlaubt". In this sentence
> "Schuh", 0 and 5 would all be replacements.

You'd need to define your issue-message as `"Das Wort {word} von Position {from} bis {to} ist unerlaubt"` and provide the replacement `{ word: "Schuh", from: 0, to: 5 }` to `addIssue`.

Korosensei42:
Everything is clear now. I approve.

PR #43, opened 2019-06-12
Ronja kaehne/improved text formatter including positioning

RonjaKaehne:
i built a second version of the text formatter using a tokenization for regular expressions. also added a setPosition method in the position.ts for easy and fast issue creating


src/issueHandling/position.ts
```diff
@@ -3,3 +3,11 @@ export default interface Position {
...
 	toIndex: number;
 }
+
+export function setPosition(from: number, to: number) {
+	const position: Position = {
+		fromIndex: from,
+		toIndex: to,
+	};
+	return position;
+}
```
Khazuar: I don't really see the added benefit in this function - it's basically the same whether I type `setPosition(5, 8)` or ` `{ fromIndex: 5, toIndex: 8 }`. 
Additionally, the name suggests you're actually changing something by calling the method, like a setter. A better name would be `createPosition` or `newPosition`.

Khazuar: This method is still not adding anything anywhere.

Korosensei42:
It does work correctly for inputs like `$a und (not a)$`. I.e. it does not display any errors. Especially it does not complain about `not` not being in the vocabulary. Judging by that, the swap within the pipeline seems to be successful.

However, in the example above nothing is displayed at all. Is this how you imagined it to be? In my opinion that's fine, but if it doesn't meet your expectation maybe something is wrong still.

Korosensei42:
Also some of the syntactical errors regarding logical operators seem to be too fine. By that I mean:
The user input `$a und (not sonne)$` returns the following errors:

```
Es fehlen logische Operatoren in der Formel.
Von Index 12 bis 13
Es fehlen logische Operatoren in der Formel.
Von Index 13 bis 14
Es fehlen logische Operatoren in der Formel.
Von Index 14 bis 15
Es fehlen logische Operatoren in der Formel.
Von Index 15 bis 16
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 13 bis 18
```

Only one of the four near-identical messages at the start should show up. Meaning, it should read
```
Es fehlen logische Operatoren in der Formel.
Von Index 13 bis 16
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 13 bis 18
```

In my opinion it could even be considered having only the last error show up in this instance, since I guess that the logical-operator-error only exists because `sonne` does not get recognized as a propositional variable. In fact, maybe we could think of some improvement like the following:

If in the user input a logic- and a vocabulary-based error happen, then throw some other kind of error like:

```
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 13 bis 18
An dieser Stelle wird eine aussagenlogische Variable erwartet.
```

PR #49, opened 2019-06-18
TimothyGillespie/integratingSWIServer

TimothyGillespie:
This will integrate diproche issue recognition.

Usage:
Run the SWIServer on port 5000

1. Find the hello.pl file.
2. Put the current diproche program into the same folder.
3. Open the hello.pl file.
4. Type in: `server(5000).`
5. Leave it open while using the webinterface.

Currently the textfield doesn't convert from normal text to the list format. This needed to put in there.
The proxy setting in the package.json is used to circumvent CORS issue. Though this can be implemented later so it can be done without the proxy. For this see:

Prolog: https://www.swi-prolog.org/pldoc/man?section=httpcors
Javascript/Typescipt: https://www.html5rocks.com/en/tutorials/cors/

The circumvented issue in question: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,5 +36,6 @@
 		"@types/react-dom": "16.8.3",
 		"shx": "^0.3.2",
 		"tslint": "^5.17.0"
-	}
+	},
+	"proxy" : "http://localhost:5000" 
 }
--- a/src/components/proofEditor/issueInformation.tsx
+++ b/src/components/proofEditor/issueInformation.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import Issue from "../../issueHandling/issue";
 import IssuePositionInformation from "./issuePositionInformation";
+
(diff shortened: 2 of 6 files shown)


TimothyGillespie:
An example of what could be entered into the textbox right now:

`[[wir,zeigen,[a,and,b],<->,[b,and,a]],[=>],[angenommen,a,and,b],[dann,a],[ferner,gilt,b],[damit,folgt,b,and,a],[qed],[abs],[<=],[nehmen,wir,jetzt,an,dass,b,and,a],[dann,folgt,a],[ausserdem,folgt,b],[also,gilt,nun,a,and,b],[qed],[damit,gilt,[a,and,b],<->,[c,and,a]]].`

PR #53, opened 2019-06-24
Ronja kaehne/second working gui

RonjaKaehne:


Khazuar:
You'll not need to put everything under `src/components/GUI`. `src/components` is already exclusive to react-components, therefore GUI-only.

src/App.tsx
```diff
@@ -1,14 +1,29 @@
 import React, { Component } from "react";
-import { ProofEditor } from "./components/proofEditor/proofEditor";
+import NavBar from "./components/GUI/mainPage/navBar";
+import SiteLayout from "./components/GUI/mainPage/siteLayout";
+import ProofEditor from "./components/GUI/proofEditor/proofEditor";
 
-class App extends Component {
+// tslint:disable-next-line: no-empty-interface
```
Nifdee: You don't need empty interfaces; imo `class App extends Component` is clearer than declaring empty interfaces and then disabling the lint rule that forbids that.

src/App.tsx
```diff
@@ -1,14 +1,29 @@
...
+export interface Props {
+}
+
+// tslint:disable-next-line: no-empty-interface
+interface State {
+}
+
+class App extends Component<Props, State> {
 	public render() {
+		window.location.href = "#startseite";
```
Nifdee: Why is this necessary?

src/components/CSS/navBar.module.scss
```diff
@@ -0,0 +1,10 @@
+li a {
```
Nifdee: This is dangerous; as far as I understand it, it'll mean that *every* link inside a `li` element will be styled as it is defined here. Depending on whether this is what you want, there are a few possible solutions:

1. explicitly add a class name to the elements you want to style, and use that class name here instead of `li a` (use this if you want to style only links that are created within the `navBar` component, because you can easily add class names to those elements)
2. wrap this in a class name - `.navBar li a {...}` (use this if you want to style all links inside `li` elements inside the `navBar` component, even when those components are created elsewhere)
3. move this declaration to an app-wide `scss` file so that it's obvious that these styles apply everywhere, not just inside the `navBar`.

In general, it's less error-prone to use custom class names directly on the elements you want to style than depending on HTML element hierarchy. The more you depend on some specific nesting, the harder it is to keep the styling and the component creating code in sync.

src/components/GUI/mainPage/navBar.tsx
```diff
@@ -0,0 +1,46 @@
...
+import NavDropdown from "react-bootstrap/NavDropdown";
+import { Props } from "../../../App";
+import navBar from "../../CSS/navBar.module.scss";
+
+interface State {
+	name: string;
+	href: string;
+}
+
+export class NavBar extends React.Component<Props, State> {
```
Nifdee: Nit: I'd use more expressive names, e.g. `NavigationBar`. It's easier to read and understand.

src/components/GUI/mainPage/navBar.tsx
```diff
@@ -0,0 +1,46 @@
...
+		href: "",
+	};
+
+	public render() {
+		// return navbar;
+		return <div className={navBar.page}>
+			<Navbar bg="transparent">
+				<Navbar.Toggle aria-controls="navbar" />
+				<Nav className="navbar">
+					<li><Nav.Link className="#mainsite" href="#startseite">Startseite</Nav.Link></li>
```
Nifdee:
```suggestion
					<li><Nav.Link className="mainsite" href="#startseite">Startseite</Nav.Link></li>
```

src/components/GUI/mainPage/siteLayout.tsx
```diff
@@ -0,0 +1,25 @@
+import "bootstrap/dist/css/bootstrap.min.css";
+import React from "react";
+import { Props } from "../../../App";
+import siteLayout from "../../CSS/siteLayout.module.scss";
+
+// tslint:disable-next-line: no-empty-interface
+interface State {
```
Nifdee: You don't need empty interfaces; simply don't declare a state type in line 10.

src/components/GUI/mainPage/siteLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export class SiteLayout extends React.Component<Props, State> {
```
Nifdee: What does this component do?

src/components/GUI/proofEditor/issuePositionInformation.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
-import Position from "../../issueHandling/position";
-import styles from "./proofEditor.module.scss";
+import Position from "../../../issueHandling/position";
+import styles from "../../CSS/proofEditor.module.scss";
```
Nifdee: style files are usually kept next to their component's `tsx` files because they are so tightly coupled. Is there a reason to move all `scss` files into a separate `CSS` folder?

TimothyGillespie:
@RonjaKaehne You could have actually continued with this PR and wouldn't need to make a new one. Since this one is not needed anymore I will close it.

PR #54, opened 2019-06-27
Implement react router and some changes

TimothyGillespie:
Implements the react router and changes some things.
It is not a clean version yet but servers a structural purpose for now.


src/App.tsx
```diff
@@ -13,17 +13,68 @@ interface State {
...
-			<ProofEditor />
-			</div>
+			<BrowserRouter>
+				<React.Fragment>
+					<Navigation />
+					<Switch>
+						<Route path="/" component={Startseite} exact />
+						<Route path="/aussagenlogisches_beweise" component={ProofEditor} exact />
+						<Route path="/einstellungen" component={Einstellungen} exact />
+						<Route path="/kontakt" component={Kontakt} exact />
```
Khazuar: It's nicer to read if you write `exact path=...` here.

src/App.tsx
```diff
@@ -13,17 +13,68 @@ interface State {
...
+	);
+};
+
+const ErrorPage404 = () => {
+	return (
+		<div>
+			<p>404 Page Not Found</p>
+		</div>
+	);
+};
```
Khazuar: These should all become their own proper components in separate modules (files)

TimothyGillespie: I put them in the file for now to avoid putting more to the folder structur which might be changed later. But this should definitly be the case later.

Nifdee:
Looks good! But "some changes" really isn't a good description for a PR :D

PR #63, opened 2019-07-01
Timothy gillespie/componentStructur

TimothyGillespie:
Externalizing the modules and simplifying the App.tsx structur by removing the Props and State interface.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1127,6 +1127,11 @@
         "@types/react-router": "*"
       }
     },
+    "@types/shortid": {
+      "version": "0.0.29",
+      "resolved": "https://registry.npmjs.org/@types/shortid/-/shortid-0.0.29.tgz",
+      "integrity": "sha1-gJPuBBam4r8qpjOBCRFLP7/6Dps="
+    },
     "@types/tapable": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/@types/tapable/-/tapable-1.0.2.tgz",
@@ -6665,8 +6670,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -6690,7 +6694,6 @@
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -6703,8 +6706,7 @@
         },
         "code-point-at": {
           "version": "1.1.0",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "concat-map": {
           "version": "0.0.1",
@@ -6713,8 +6715,7 @@
         },
         "console-control-strings": {
           "version": "1.1.0",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -6817,8 +6818,7 @@
         },
         "inherits": {
           "version": "2.0.3",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "ini": {
           "version": "1.3.5",
@@ -6828,7 +6828,6 @@
         "is-fullwidth-code-point": {
           "version": "1.0.0",
           "bundled": true,
-          "optional": true,
           "requires": {
             "number-is-nan": "^1.0.0"
           }
@@ -6841,20 +6840,17 @@
         "minimatch": {
           "version": "3.0.4",
           "bundled": true,
-          "optional": true,
           "requires": {
             "brace-expansion": "^1.1.7"
           }
         },
         "minimist": {
           "version": "0.0.8",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "minipass": {
           "version": "2.2.4",
           "bundled": true,
-          "optional": true,
           "requires": {
             "safe-buffer": "^5.1.1",
             "yallist": "^3.0.0"
@@ -6871,7 +6867,6 @@
         "mkdirp": {
           "version": "0.5.1",
           "bundled": true,
-          "optional": true,
           "requires": {
             "minimist": "0.0.8"
           }
@@ -6944,8 +6939,7 @@
         },
         "number-is-nan": {
           "version": "1.0.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "object-assign": {
           "version": "4.1.1",
@@ -6955,7 +6949,6 @@
         "once": {
           "version": "1.4.0",
           "bundled": true,
-          "optional": true,
           "requires": {
             "wrappy": "1"
           }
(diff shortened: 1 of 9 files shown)


TimothyGillespie:
Should actually have been going into #62 but it has been merged.

src/App.tsx
```diff
@@ -1,19 +1,17 @@
 import React, { Component } from "react";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 
+// tslint:disable: ordered-imports
```
Korosensei42: What is the reason you disabled the ordered-imports?

src/App.tsx
```diff
@@ -1,19 +1,17 @@
...
 import MainPage from "./components/mainPage/mainPage";
 import Navigation from "./components/navigationBar/navigationBar";
 import ProofEditor from "./components/proofEditor/proofEditor";
+import Einstellungen from "./components/einstellungen/einstellungen";
```
Korosensei42: Internally in our programfiles the variables and imports should be consistently in english in my opinion.
Even if the displayed text later is in german.

package-lock.json
```diff
@@ -6665,8 +6670,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
```
Korosensei42: Out of curiosity: What is the effect of removing the optional predicate here and in the other functions?

src/components/proofEditor/issueInformation.tsx
```diff
@@ -7,14 +8,18 @@ export interface Props {
 	issue: Issue;
 }
 
-export class IssueInformation extends React.Component<Props> {
+export class IssueInformation extends React.Component<Props, {}> {
 
 	public render() {
```
Korosensei42: I would wish for at least one or two lines of comment explaining what's happens exactly in this rendermethod

src/components/proofEditor/issueInformation.tsx
```diff
@@ -7,14 +8,18 @@ export interface Props {
...
 
 		let displayedPosition: any = "";
 		const position = this.props.issue.position;
 		if (position) {
```
Korosensei42: `position` is a weird name for a boolean. Also the fact that you set it equal to a position of an issue is confusing. Additionally it is not obvious for me what "props" stands for. Properties maybe? If so: Properties of what?

src/components/proofEditor/issueInformation.tsx
```diff
@@ -7,14 +8,18 @@ export interface Props {
...
-			displayedPosition = <IssuePositionInformation position={position!} />;
```
Korosensei42: Is position now a boolean or an IssuePositionInformation?

Also; What does `position!` do? I know `!position` which negates a boolean expression.
Does it do the same in the postfix notation?

src/App.tsx
```diff
@@ -1,19 +1,17 @@
...
+import Kontakt from "./components/kontakt/kontakt";
```
Korosensei42: What is the difference between `Kontakt` and `Impressum`?
Online I found this http://www.impressum-recht.de/impressum-kontakt-angaben.html.

At the bottom of the page it says `als Bezeichnung bieten sich "Impressum" oder "Kontakt" an;`,
so it seems like these two words are synonyms.

src/App.tsx
```diff
@@ -1,19 +1,17 @@
 import React, { Component } from "react";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 
+// tslint:disable: ordered-imports
 import MainPage from "./components/mainPage/mainPage";
 import Navigation from "./components/navigationBar/navigationBar";
```
Korosensei42: What do you expect when you click on `Navigation`? I don't have any imagination here.

src/App.tsx
RonjaKaehne: I think you should see them as dummy until we made a impressum or contact page. Actually you can see how the menu works. We can add the impressum/contact page later but in this version we should focus of the functionality of the basic elements.

src/App.tsx
RonjaKaehne: Navigation is a class which includes the elements from the navigation bar. for example if you want to use a sidebar instead to have to modify this class. look inside ./components/navigationBar/navigationBar to see how it works or just do "npm run start" on this branch

Korosensei42: "npm run start" doesn't work for me. I get this error
```
./src/components/proofEditor/issueInformation.tsx
Module not found: Can't resolve 'shortid' in 'C:\Users\kweir\Desktop\Uni\SoSe19 - Vorlesungen_Arbeit\Diproche\github\webinterface\src\components\proofEditor'
```

A git pull didn't resolve this problem either. That's why I asked in the first place. Do you have another suggestion for a fix?

RonjaKaehne: Did u tried "npm install" first? the branch worked for me after doing that.

Korosensei42: No, I didn't. That seemed to be the problem for me, too. Thanks!

src/App.tsx
TimothyGillespie: I intend the orders of the imports to follow the order of the navigation/react-router listing so you can find the specific component easier.

src/App.tsx
TimothyGillespie: Agreed. Though I would like to see the link reflect the component name.

package-lock.json
TimothyGillespie: I am not sure. This is just changed by npm automatically and I just commit it. I googled if I should commit it, but apparently I should.

src/components/proofEditor/issueInformation.tsx
TimothyGillespie: The ! after the variable/function etc. means that this variable/return won't be null or undefined. This is for the typescript compiler since it sometimes can be impossible for an object to become undefined, but the compiler won't recognize it and throw an error.

src/components/proofEditor/issueInformation.tsx
TimothyGillespie: Now that you understand the concepts used here do you think it is still necessary?

src/App.tsx
TimothyGillespie: I just implemented what was given before so no design choice on my side. Though I think sometimes there can be an impressum and a contact page and the impressum would hold the contact information while the contact page has a formular which will send an email to the site owner or support.

But we will see what we need later and it shouldn't be hard to change it. So as @RonjaKaehne says they are pretty much dummys for now.

src/components/proofEditor/issueInformation.tsx
TimothyGillespie: I forgot to address the props question.
`props` is the standard attribute of a react component. So it should be clear when I pass the interface with the name Props. When you pass the interface into React.Component<Props, State> like this then Props will become the definition of the props and State of the states. In this case we don't use states (so I could have and should have built a stateless component) so therefore it's just {} for the State.

When you create a react component you will use them in a different location like this:

<ComponentName />

But you can also pass information to it like this.
<ComponentName value="Some String" />

You can then access this information value like via:
`this.props.value`

The interface that is passed is for Typescript to have a description of the props. It just mirrors the internal attribute name.

TimothyGillespie:
@Korosensei42 I haven't made any changes, but I have addressed your questions. Does this implementation make more sense to you now?

src/App.tsx
Korosensei42: Okay, this is a reason I can agree on

Korosensei42:
@TimothyGillespie I think everything is clear now. Thanks!

Issue #64, opened 2019-07-01
The vocabulary check is case sensitive


TimothyGillespie:
### Expected Behavior:
The input `Es sei $A und B$` returns no error. Generally, allowed words will still be allowed even if they the first letter is capitalized.

### Current Behavior
The input `Es sei $A und B$` returns three errors, while `es sei $a und b$` returns none.

### Steps to reproduce
1. Go to Spielweise -> Aussagenlogisches Beweisen
2. Enter  `Es sei $A und B$` and `es sei $a und b$`

### Possible Implementation
The text, when received as a parameter, could be converted to all lowercase.

PR #65, opened 2019-07-02
Make vocabChecker case-unsensetive

Korosensei42:
I might have pushed something onto your branch @TimothyGillespie, when actually I wanted to just make a PR. Sorry for that! Anyways, my change should only affect the case sensitiveness. It should handle the two scenarios you described in #64 correctly now.


TimothyGillespie:
@Korosensei42 Thanks for the fix. Next time there is a fix like this you can merge it directly into the master branch (with the master as a base branch of course).
The fix is independet of the JSDocs and it can cause some problems. For example when the JSDocs is being refused in the end the change will never find it's way into master and it is not immediately available to every branch.

But we can just merge it into the JSDocs for now.

PR #68, opened 2019-07-02
RonjaKaehne/fixingSplittingBugOnIssue#60

### LINKED ISSUE #60: bug by splitting the user input
usingmultiple seperator next to each user actually throw bugs.
example $a and b$.
"$a and b" should get splitted and the part bewteen the second "$" and the "." but there is no space to split. this needs to get catched.

RonjaKaehne:
Fixes #60

### DIFF
--- a/src/input_formatter/text_formatter.test.tsx
+++ b/src/input_formatter/text_formatter.test.tsx
@@ -92,13 +92,13 @@ test("replace detected expression-elements into readable prolog commands", () =>
 
 test("expression formatter testcase", () => {
 	const result: string = textFormatter("Just another test $a .and a$.. test $test2$$test3$");
-	const expectedResult = "[[just,another,test],[a .,and,a],[test],[test2],[test3]].";
+	const expectedResult = "[[just,another,test],[a .,and,a],[test],[test2],[test3]]";
 	expect(result).toEqual(expectedResult);
 });
 
 test("expression formatter testcase", () => {
 	const result: string = textFormatter("Es gilt: $((a&b)<->(b&a))$");
-	const expectedResult = "[[es,gilt],[[[a,and,b],<->,[b,and,a]]]].";
+	const expectedResult = "[[es,gilt],[[[a,and,b],<->,[b,and,a]]]]";
 	expect(result).toEqual(expectedResult);
 });
 
@@ -110,7 +110,7 @@ test("splitting full text into readable prolog format including formatted expres
 	const expectedResult = "[[hello,this,is,a,test],[is,it,working]," +
 		"[i,hope,so],[abs],[paragraphs,are,marked,with,an,abs,list]," +
 		"[abs],[abs],[abs],[here,is,also,an,expression],[[a,and,[b,<->],->," +
-		"d,or,neg,e]],[and,a,second,one],[[5,+,12,=,3,*,5,+,2]]].";
+		"d,or,neg,e]],[and,a,second,one],[[5,+,12,=,3,*,5,+,2]]]";
 	expect(result).toEqual(expectedResult);
 	const issue = listAllIssues().find(i => i.code === "MISSING_STATEMENT_INSIDE");
 	expect(issue).toEqual({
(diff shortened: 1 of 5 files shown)


TimothyGillespie:
If you put the issue in the description it will show up in the issue #60 too and the PR and the two are linked then.

src/input_formatter/text_formatter.test.tsx
```diff
@@ -91,10 +91,17 @@ test("replace detected expression-elements into readable prolog commands", () =>
 	expect(result).toEqual(expectedResult);
 });
 
+test("expression formatter testcase", () => {
+	emptyIssueList();
```
TimothyGillespie: You keep repeating `emptyIssueList()` in every test. The operations you repeat before every test can be perfectly be put into the set up. See the beforeEach function here: https://jestjs.io/docs/en/setup-teardown

src/input_formatter/text_formatter.test.tsx
```diff
@@ -105,9 +112,9 @@ test("splitting full text into readable prolog format including formatted expres
 		"abs List: \n \n \n Here is also an expression: $[AUNDbracketLEFTB<-->]-> D ODERNOT E]$ " +
 		"And a second one: $[5 ADD 12 equal 3 mal 5 plus 2]$");
 	const expectedResult = "[[hello,this,is,a,test],[is,it,working]," +
-		"[i,hope,so],[abs],[paragraphs,are,marked,with,an,abs,list:]," +
-		"[abs],[abs],[abs],[here,is,also,an,expression:],[[a,and,[b,<->],->," +
-		"d,or,neg,e]],[and,a,second,one:],[[5,+,12,=,3,*,5,+,2]]].";
+		"[i,hope,so],[abs],[paragraphs,are,marked,with,an,abs,list]," +
+		"[abs],[abs],[abs],[here,is,also,an,expression],[[a,and,[b,<->],->," +
+		"d,or,neg,e]],[and,a,second,one],[[5,+,12,=,3,*,5,+,2]]].";
```
TimothyGillespie: I realized it just now. The string will always end with a `.` which is because I specified it needs to end on that for the SWI-Server. But I think this will be better put into the code that makes a request to the SWI Server. I will add this later, but this can be put away for now.

src/input_formatter/text_formatter.test.tsx
```diff
@@ -91,10 +91,17 @@ test("replace detected expression-elements into readable prolog commands", () =>
 	expect(result).toEqual(expectedResult);
 });
 
+test("expression formatter testcase", () => {
+	emptyIssueList();
+	const result: string = textFormatter("Just another test $a .and a$.. test $test2$$test3$");
+	const expectedResult = "[[just,another,test],[a .,and,a],[test],[test2],[test3]].";
```
TimothyGillespie: If I understand this correctly the dot will only appear in the list if it put inside of expressions marked by `$...$`?

RonjaKaehne: yes, like it should be.

src/input_formatter/text_formatter.test.tsx
RonjaKaehne: good idea. thx. i changed it

src/input_formatter/text_formatter.ts
```diff
@@ -9,7 +9,8 @@ import { Regexes } from "./regexes";
  * formatted expressions, list of words (setences) and paragraph marker (empty Array)
  */
 export function textFormatter(input: string): string {
-	const splittedText = input.split(Regexes.inputSeparator);
+	let splittedText = input.split(Regexes.inputSeparator);
+	splittedText = splittedText.filter(x => x !== undefined && x !== null && x !== "");
```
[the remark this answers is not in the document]
RonjaKaehne: this doesn't work and throws errors like this:  "Cannot read property 'length' of undefined". i think the code doesn't need to be changed here.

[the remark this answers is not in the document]
TimothyGillespie: My bad. It should have been just x => `x`, not `!x`. I'll change the suggestion above.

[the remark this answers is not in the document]
RonjaKaehne: okay this works for me. thx for your advice

src/input_formatter/text_formatter.test.tsx
RonjaKaehne: okay i will remove adding the dot at the end

RonjaKaehne:
bug is fixed and also some minor code improvements are made

Issue #50, opened 2019-06-20
Fix indexing for vocabErrors


Korosensei42:
Currently inputs like `$sonne und sonne$` create a duplicate of just a single issue, namely
```
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 1 bis 6
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 1 bis 6
```

This should be fixed, meaning it should return issues in the form of
```
Das Wort 'sonne' in der Eingabe ist nicht erlaubt.
Von Index 1 bis 5 und von Index 11 bis 15
```

Korosensei42:
@TimothyGillespie @RonjaKaehne This bug is currently not visible on master, since we didn't push an update regarding the GUI to this branch recently. In which of your branches is this bug visibly present?

TimothyGillespie:
It is reproducable on master for me. What Browser are you using and what Operating System?

Korosensei42:
I am using Firefox on Windows 10

TimothyGillespie:
Hmm that's the same as mine. Did you git pull?

Korosensei42:
I think I did so, but I will try it again.

Korosensei42:
Okay, it works now. I don't know why it didn't earlier. Maybe I really forgot to pull, even though I was certain I did so

Korosensei42:
A partial fix is available in #74

TimothyGillespie:
This should be completely solved now with #74 as far as I can see?

Korosensei42:
This is indeed solved, yes. Thanks for closing it. Forgot about that

RonjaKaehne:
we still got a bug here but another one now:
for example:
ffff ff
returns the following issues:
Das Wort 'ffff' in der Eingabe ist nicht erlaubt.
Von Index 0 bis 4
Das Wort 'ff' in der Eingabe ist nicht erlaubt.
Von Index 0 bis 2
Das Wort 'ff' in der Eingabe ist nicht erlaubt.
Von Index 1 bis 3
Das Wort 'ff' in der Eingabe ist nicht erlaubt.
Von Index 2 bis 4
Das Wort 'ff' in der Eingabe ist nicht erlaubt.
Von Index 5 bis 7

But only the first and the last issue should get collected here

RonjaKaehne:
i guess we can catch this if we search for white spaces between words first.

PR #74, opened 2019-07-04
Fix indexing error

Korosensei42:
The proposed change fixes the indexing error for inputs like
`A und Sonne und Sonne` described in #50.

It still prints out a duplicate of the same issue, but
the positioning works fine now.

### DIFF
--- a/src/vocabularyChecker/detectWrongSyntax.ts
+++ b/src/vocabularyChecker/detectWrongSyntax.ts
@@ -20,14 +20,28 @@ export function collectInvalidWordsInIssues(text: string) {
 	const positions: Position[] = [];
 	for (const word of invalidWords) {
 		if (!(allowedWords.includes(word.toLowerCase()))) {
-			const temp: Position = {
-				fromIndex: text.indexOf(word),
-				toIndex: (text.indexOf(word) + word.length),
-			};
-			positions.push(temp);
-			addIssue("INVALID_WORD", temp, {word});
+			const allPositionsOfInvalidWord = getPositionsOfInvalidWord(word, text);
+			for (const pos of allPositionsOfInvalidWord) {
+				positions.push(pos);
+				addIssue("INVALID_WORD", pos, {word});
+			}
 		}
 	}
+}
+
+export function getPositionsOfInvalidWord(invalidWord: string, text: string): Position[] {
+	let tempText = text;
+	const result: Position[] = [];
+	let offSet = 0;
+	while (tempText.includes(invalidWord)) {
+		const tempPos: Position = {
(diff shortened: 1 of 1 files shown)


src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -20,16 +20,30 @@ export function collectInvalidWordsInIssues(text: string) {
...
 		if (!(allowedWords.includes(word.toLowerCase()))) {
-			const temp: Position = {
-				fromIndex: text.indexOf(word),
-				toIndex: (text.indexOf(word) + word.length),
-			};
-			positions.push(temp);
-			addIssue("INVALID_WORD", temp, {word});
+			const allPositionsOfInvalidWord = getPositionsOfInvalidWord(word, text);
+			for (const pos of allPositionsOfInvalidWord) {
+				positions.push(pos);
```
TimothyGillespie: As far as I can tell the variable position and therefore this push doesn't do anything (anymore)?

src/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -20,16 +20,30 @@ export function collectInvalidWordsInIssues(text: string) {
...
+				addIssue("INVALID_WORD", pos, {word});
+			}
 		}
 	}
 }
 
+export function getPositionsOfInvalidWord(invalidWord: string, text: string): Position[] {
```
TimothyGillespie: You could simplify this function by knowing that the `String.prototype.indexOf` method can take a second parameter to start searching at. I created pull request #76 to show what it could look like.

I think exported Functions should always have a JSDoc. If I am missing them somewhere then feel free to point it out to me. It is the easiest to create them right away when creating the function. I regretted having pushed the JSDocs to a later time for a seperate pull request.

Korosensei42: You are right. I will add it once I get to it

src/vocabularyChecker/detectWrongSyntax.ts
Korosensei42: I will have to test this, but I think you are right, if I am not mistaken.

Korosensei42:
@TimothyGillespie I adressed your feedback. Your suggestion for `const invalidWords = getInvalidWords(text);` was on point. I also included your suggestion to simplify `getPositionsOfInvalidWord`.
Thank you very much!

I will just add some documentation right now, since it is pretty sparse in this file.

Korosensei42:
From a functional point of view this should be ready to be merged now. Feel free to leave any suggestions, though.

Issue #69, opened 2019-07-03
Adding a textbox with autosize


RonjaKaehne:
The input textbox for the profchecker should automaticly growing if it's needed.

TimothyGillespie:
This issue will very likely be solved by my adjustments as a by product since moving from a textarea to an editable div seems the most practical.

RonjaKaehne:
okay then i assign this issue to you.

TimothyGillespie:
Pretty much solved with #75 automatically.

Issue #19, opened 2019-05-09
Present issues directly in the input


Khazuar:
E.g. by underlining the appropriate text-passages in red. More information should be shown when hovering over the text. Probably depends on #16

PR #75, opened 2019-07-04
Show errors in the user input

TimothyGillespie:
Referring to #19.
Solves #69 as a side effect.

The general basis is set up now. There is no marking up in text. This is missing still.
Currently proofs are checked on every change. This can be fetched with the "onKeyDown" event (ContentEditable doesn't support many events) so that it only get's checked when a certain character like a space or a "." is pressed. This is up to debate.

For implementing the mark up in the user input directly I only need to find a good way to put HTML tags in multiple positions since the index would change and this would invalidate the given position from the issue.


TimothyGillespie:
There are still some known issues. Tests do exist for addStringIgnoringHTML, but since they still fail they are not yet commited. 

Two very severe issues are that putting something like "3 < 5" will break addStringIgnoringHTML since it uses a rather naive approach to recognize HTML tags.

The second severe issue is that the proofEditor is hardly editable because of the way it is being rendered and updated. Also: Deleting a letter from a marked word will make the last letter of the marked word appear outside of the marking.

I will now work on fixing this bugs.

TimothyGillespie:
addHTMLToString has one bug still. It has issues, when the string begins with an HTML Tag it seems as it doesn't count this one in then.

It was also intended to add text inside of HTML tags, but thinking about the later use case it seems to be more useful to actually place them outside of the ignored HTML tags. So the test will have to be changed to this.

TimothyGillespie:
The main issues has been, that I assumed the Array.prototype.map method- unlike the others - would mutate the array. But it didn't.

TimothyGillespie:
I am currently stuck. When you go to the proofEditor and just type some things and add [enters] it will get messed up. It will update the issues on pressing space. I don't see where the issue with the code is and it will try to insert <mark> into <br> and the like.
Maybe someone else can spot the issue which I am just not seeing.

TimothyGillespie:
Why can you not the test the branch with the merge conflicts? Ideally it would be up to date with the master branch but implemented functionality is still there. I will resolve the merge conflicts tomorrow though now that the big changes are through.

Korosensei42:
Oh, only at a second glance I realized I don't have react-contenteditable installed. Gonna do that and then try it out

Korosensei42:
It appears like there is still at least this new error:
https://gyazo.com/4bc3b33ff3f1203ccb01a7c5b1692bc4

Maybe it will be resolved when the conflicts github detected are resolved

TimothyGillespie:
It works properly on my machine. I cannot reproduce the error or wouldn't know how. It's a type error. Generally speaking such an error can't occure because of merge conflicts. Hmm.
Can you describe how you get this error?

TimothyGillespie:
Apparently when I merged it and undid the merge it didn't realize I was still resolving merge conflicts. The merge conflicts are resolved now. The proofEditor of this branch is yet to be debugged in this implementation though.

Korosensei42:
> It works properly on my machine. I cannot reproduce the error or wouldn't know how. It's a type error. Generally speaking such an error can't occure because of merge conflicts. Hmm.
Can you describe how you get this error?

I checked out your branch, pulled it and tried to `npm start` via VS-Code. In the brwoser the first of these type errors was displayed. Then I checked the file it told me where the error should be and there were three of them. Not I tried again and apparently there are five now.
They all are in proodEditor.tsx. Specifically there are problems with the variables `ref` (in line 36), `onChange` (in line 39) and `onKeyDown` (in line 40) and there are problems with `checkInput` (in line 120) and `this.props.userInput` (in line 121).
If you want a detailled error description, I can provide it.
If you want to, I can give you each error description individually.

TimothyGillespie:
I have made some changes. Could you git pull again and tell me if the same errors occure?

Korosensei42:
I mistyped, sorry!

By 

> Not I tried again and apparently there are five now.

I meant "Now I tried again and apparently there are five now."
So I already pulled and checked again.

TimothyGillespie:
That seems odd then. I assume the error is now in the proofEditor/proofEditor.tsx file?
The odd thing is that they don't appear for me at all. I will see what I will do about this. Eventually I will make a new branch from master and implement it into the new system right away. Might be easier and quicker on two fronts now.

PR #77, opened 2019-07-07
Add tslint rules for file casing and jsdocs

TimothyGillespie:
Adding tslint rules for consistency and enforcing documentation.
The rules are of course up to debate, but this is my suggestion. Plese voice your opinion.

If we decided on them: Fixes to fit the new rules should be put in a new PR and merged into this one.

A link with the tslint options can be found here: https://palantir.github.io/tslint/rules/


tslint.json
```diff
@@ -14,7 +14,16 @@
 			2
 		],
 		"no-conditional-assignment": false,
-		"arrow-parens": false
+		"arrow-parens": false,
+		"no-redundant-jsdoc": true,
```
TimothyGillespie: Disallows defining the types in the JSDocs when typescript specified them.

tslint.json
```diff
@@ -14,7 +14,16 @@
...
-		"arrow-parens": false
+		"arrow-parens": false,
+		"no-redundant-jsdoc": true,
+		"completed-docs": [
+		  true,
+		  {
+		    "functions": {"visibilities": ["exported"]},
+		    "methods": {"privacies": ["public", "protected"]}
+		  }
+		],
```
TimothyGillespie: Forces to have JSDocs for all exported functions as well as public and protected methods.

tslint.json
```diff
@@ -14,7 +14,16 @@
...
+		"file-name-casing": [true, "camel-case"]
```
TimothyGillespie: Forces to have .ts and .tsx files in camelCases which is pretty much the convention that we have now.

src/App.tsx
```diff
@@ -1,3 +1,4 @@
+// tslint:disable:file-name-casing
```
Nifdee: Is it not possible to just rename this file, too?

src/inputFormatter/expressionFormatter.ts
```diff
@@ -70,13 +73,21 @@ export function replaceASingleExpressionElementIntoPrologCode(preformattedExpres
 		;
 }
 
+/**
+ * This function just calls all the other function with the abilaty to detect wrong user input.
```
Nifdee:
```suggestion
 * This function just calls all the other function with the ability to detect wrong user input.
```

src/inputFormatter/expressionFormatter.ts
```diff
@@ -26,6 +26,9 @@ export function preFormatExpressionFromImput(expression: string): string[] {
 }
 
 /**
+ * format the different
+ * input styles for logical vocabulary into one single style supported by SWI prolog
+ * @param the expression that could still including different input styles
```
Nifdee:
```suggestion
 * @param the expression that could still include different input styles
```

src/inputFormatter/expressionFormatter.ts
```diff
@@ -70,13 +73,21 @@ export function replaceASingleExpressionElementIntoPrologCode(preformattedExpres
...
+ * @param expressionPosition where to match issueCodes with the right position
+ */
 export function expressionIssueDetector(preFormattedExpression: string[], expressionPosition: number) {
 	detectBracketIssues(preFormattedExpression, expressionPosition);
 	detectMissingStatementsOrConnector(preFormattedExpression, expressionPosition);
 }
 
 /**
- * Check the expression for an equal amount of opened and closed brackets.; it can detect wrong input
+ * Check the expression for an equal amount of opened find closed brackets.
```
Nifdee: This description suggests that `)foo(` would also count as valid because it's the same number of open and closed brackets. I'd just say something like "Check if each opening bracket is closed correctly and vice versa"

src/inputFormatter/expressionFormatter.ts
```diff
@@ -108,7 +119,10 @@ export function detectBracketIssues(expression: string[], expressionPosition: nu
 }
 
 /**
- * check the expression for correctness that no logic operators or statements is forgotten; it can detect wrong input
+ * Check the expression for correctness that no logic operators or statements is forgotten.
+ * It can detect wrong user input and push issueCodes if it's needed.
```
Nifdee:
```suggestion
 * It can detect wrong user input and push issueCodes if necessary.
```

src/util/proofChecker.ts
```diff
@@ -7,15 +7,20 @@ const severities: string[] = ["FATALERROR", "ERROR", "WARNING", "HINT"];
...
+	* @return The syntatical, sementatical and technical issues for the given user input
 	*/
 export async function checkProof(userInput: string): Promise<readonly Issue[]> {
 	emptyIssueList();
 	createErrors(userInput);
 	return orderIssuesBySeverity(listAllIssues());
 }
 
+/**
+	* Receives the diproche reponse for a given input
```
Nifdee:
```suggestion
	* Receives the diproche response for a given input
```

src/App.tsx
TimothyGillespie: There is some odd behavior when renaming this file.
The imports of this file insist that it is capitalized. Using lower case will result in an error when importing in the index.ts file. Though it resolves the TSLint error.

So I guess the decision is either between having the consistency with the file names or consistency with the imports. 

Since a wrong import is a more fatal error I chose to keep the import consistent; in case some future update will break everything.

Nifdee: Hm, interesting... yeah in that case probably better to leave this way.

Korosensei42:
On my local machine `npm run lint` does not return any issues anymore and all tests work. However, semaphoreci says there are issues. The ones it mentions in the webbrowser are fixed on my machine and git also says that everything is up to date, i.e. the remote branch and my local one are identical.

Khazuar:
@Korosensei42: Well, on github the files that semaphore complains about are in fact still not camel-case:
camelCase

Even if they are fixed in your local checkout (`alphabetize.py` and `iterators.ts`), git does not catch on to those changes, because it's by default case-insensitive. To it, the files just seem unchanged. You will need to explicitly tell it to move these files in its internal index by using [`git mv`](https://git-scm.com/docs/git-mv).

Korosensei42:
Thanks @Khazuar. I needed a whacky workaround on the way, but now it seems fine. At least semaphoreci approves.

Khazuar:
Now it's only the conflicts that remain to be solved :)

Korosensei42:
I resolved most of the conflicts. The problem now seems to be some test from `textFormatter.test.tsx` not working and I can't make out why it doesn't work. Could someone have a look at this?

TimothyGillespie:
Will do a little later.

Korosensei42:
I correct myself. Found the problem. While trying to fix the other problems it seems like I screwed over the folder structure and / or got some duplicate files in there. I will try and fix this now. The tests work again, though

Korosensei42:
Alright, that should do it. Someone should also have a look at it though. There was a lot of trouble with the filenames, especially the ones using snake_case and DromedarCase. Due to git not behaving as I wanted I often needed to delete some file just to add it again with the proper name. I wouldn't be surprised if I forgot to actually paste back the contents of some file, leaving it empty. So especially keep your eyes open for such mistakes.

TimothyGillespie:
P.S. I checked the changes and compared it with the master branch. I don't see any removed code. Good job.

TimothyGillespie:
I did find one mistake. there is an additional vocabChecker folder with the iterator inside. This one seemed to have been left over though it seems to behave weirdly? Anyway that should maybe be taken care of as a last issue for this PR.

Korosensei42:
> I checked the changes and compared it with the master branch. I don't see any removed code. 

Thank you!

> I cannot approve it as the author of the PR. [...] It's an okay for me. I don't wanna delay the merge to avoid future conflicts.

I wanted to wait for someone's (verbal) approval.

> I did find one mistake. there is an additional vocabChecker folder with the iterator inside. This one seemed to have been left over though it seems to behave weirdly? Anyway that should maybe be taken care of as a last issue for this PR.

I just removed it. I will go ahead and approve this PR then and merge it

Korosensei42:
@TimothyGillespie I leave it up to you to delete or keep this branch, as it is your's.

PR #78, opened 2019-07-09
Ronja kaehne/inner structur for propositional logic

RonjaKaehne:
adding inner structur for propositional logic including navbar, tutorialpage, exercise page and example page


src/App.tsx
```diff
@@ -1,15 +1,13 @@
 import React, { Component } from "react";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 
-// tslint:disable: ordered-imports
```
TimothyGillespie: Why are you removing this?

src/components/propositionalLogic/propositionalLogic.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+// tslint:disable-next-line: no-empty-interface
+interface State {
+}
+
+class PropositionalLogic extends Component<Props, State> {
+	public render() {
+
+		return (
+			<BrowserRouter>
```
TimothyGillespie: You don't need the BrowserRouter here because all of this will inside the `App.tsx`'s BrowserRouter.

src/components/propositionalLogic/propositionalLogic.tsx
```diff
@@ -0,0 +1,49 @@
...
+				<React.Fragment>
+					<NavigationPropositionalLogic />
+					<Switch>
+						<Route exact path="/" component={NavigationPropositionalLogic} />
+						<Route exact path="/aussagenlogisches_beweisen/tutorial" component={TutorialPropositionalLogic} />
+						<Route exact path="/aussagenlogisches_beweisen/beispiele" component={ExamplesPropositionalLogic} />
+						<Route exact path="/aussagenlogisches_beweisen/uebungen" component={ExercisesPropositionalLogic} />
+						<Route exact path="/aussagenlogisches_beweisen/sandbox" component={FreeProverPropositionalLogic} />
+						<Route component={ErrorPage404} />
```
TimothyGillespie: This shouldn't be needed since it will fetch the defintion of the `App.tsx` then.

src/components/propositionalLogic/propositionalLogic.tsx
```diff
@@ -0,0 +1,49 @@
...
+					</Switch>
+				</React.Fragment>
+			</BrowserRouter>
+		);
+
+	}
+
+}
+
+const ErrorPage404 = () => {
```
TimothyGillespie: We have a ErrorPage404 component. If you need to use it it should be imported, otherwise we have to handle two different components which shall be the same anyway.

src/components/propositionalLogic/propositionalLogic.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+class PropositionalLogic extends Component<Props, State> {
+	public render() {
+
+		return (
+			<BrowserRouter>
+				<React.Fragment>
+					<NavigationPropositionalLogic />
+					<Switch>
```
TimothyGillespie: The Switch shouldn't be needed by the same logic.

src/App.tsx
RonjaKaehne: why not? i think we should avoid tslint disable comments in our code so just order the imports instead

src/components/propositionalLogic/propositionalLogic.tsx
RonjaKaehne: i got it. thx i rebuild it inside the App.tsx and deleted the whole propositionalLogic.tsx

src/App.tsx
TimothyGillespie: TSLint is supposed to increase readability and enforce standards. I purposefully put it there so the order of the imports will match the order of the Routes, which makes it easier to pinpoint imports. Alphabetical ordering doesn't help here.

Korosensei42:
I think we should talk about an integrated wiki within the system later. I.e.:

- Should the wiki within the webinterface just be the same as on our GitHub? (Most likely not, since on GitHub it will feature some technichalities that are not relevant to a user)
- Should it be just a collection of entries that are relevant to the user, but already exist on our GitHub wiki? 
- Should it be on the same level as our GitHub wiki or should it be tailored more towards a new user that does not know much about the fundamentals of Mathematics or Programming?

RonjaKaehne:
> 
> 
> I think we should talk about an integrated wiki within the system later. I.e.:
> 
>     * Should the wiki within the webinterface just be the same as on our GitHub? (Most likely not, since on GitHub it will feature some technichalities that are not relevant to a user)
> 
>     * Should it be just a collection of entries that are relevant to the user, but already exist on our GitHub wiki?
> 
>     * Should it be on the same level as our GitHub wiki or should it be tailored more towards a new user that does not know much about the fundamentals of Mathematics or Programming?

This is not for developer, it's for the user. Inside this wiki we can store some fundamentals of mathematics like u said or links where they can read about the theory. We also can give references like a recommended writing style or a list of allowed words. at least we can remove this wiki easy if we are not gonna use it but i wanted to show how the final structur can look like

Korosensei42:
I like the idea itself very much. Just wanted to point out, that we should clarify the details later on.

PR #79, opened 2019-07-11
RonjaKaehne/exerciseGenerator

RonjaKaehne:


TimothyGillespie:
I think we said we wanted to implement the generator later, or am I mistaken on that? Is there a corresponding issue?

TimothyGillespie:
Since it is not relevant as of now I will close it. It can be reopened later on.

PR #83, opened 2019-07-18
Fix indexing error

### LINKED ISSUE #50: Fix indexing for vocabErrors

Korosensei42:
This should fix #50.

The issue was, that an input like "ff ffff" spat out four issues when only two shoul be created.

The reason of this was, that "ff" was detected as an invalid word and the whole input was searched for additional occurences of "ff". 

The solution I implemented was: A word is classified as some sequence of characters between whitespaces. Therefore I now take the user input and append a Whitespace in the front and back of it and only look for occurences of " ff ", i.e. occurences of "ff" that have a whitespace in the front and back now. This solution is a bit hacky, but it should be justified, since it builds upon the definition of a word.

### DIFF
--- a/src/util/vocabularyChecker/detectWrongSyntax.ts
+++ b/src/util/vocabularyChecker/detectWrongSyntax.ts
@@ -2,7 +2,7 @@ import Issue from "../issueHandling/issue";
 import { addIssue } from "../issueHandling/issueMapping";
 import issueJson from "../issueHandling/knownIssues.json";
 import json from "./allowedVocab.json";
-import uniqueValues from "./iterators";
+import uniqueValues from "./Iterators";
 
 /**
  * Regular expression that looks for any and all words. A word is defined as some sequence of
--- a/src/util/vocabularyChecker/detectWrongSyntax.ts
+++ b/src/util/vocabularyChecker/detectWrongSyntax.ts
@@ -2,7 +2,7 @@ import Issue from "../issueHandling/issue";
 import { addIssue } from "../issueHandling/issueMapping";
 import issueJson from "../issueHandling/knownIssues.json";
 import json from "./allowedVocab.json";
-import uniqueValues from "./Iterators";
+import uniqueValues from "./iterators";
 
 /**
  * Regular expression that looks for any and all words. A word is defined as some sequence of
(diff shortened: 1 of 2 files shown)


src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -15,10 +15,14 @@ export const anyWord = new RegExp(/\b\w+\b/g);
 const allowedWords = json;
 export interface Position { fromIndex: number; toIndex: number; }
 
-export function collectInvalidWordsInIssues(text: string) {
+/**
+ * Adds an Issue for every invalid word.
+ * @param text the userinput
```
TimothyGillespie: The `-` is missing here.

src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -15,10 +15,14 @@ export const anyWord = new RegExp(/\b\w+\b/g);
...
+ */
+export function collectInvalidWordsInIssues(text: string): void {
 	const invalidWords = getInvalidWords(text);
 	for (const word of invalidWords) {
-		if (!(allowedWords.includes(word.toLowerCase()))) {
+		if ((!(allowedWords.includes(word.toLowerCase())))) {
```
TimothyGillespie: Is this extra parenthesis necessary for this to work? Otherwise I think the extra paranthesis makes it less readable.

src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -44,18 +54,18 @@ export function getPositionsOfInvalidWord(invalidWord: string, text: string): Po
 }
 
 /**
- * Bundling the functionalities of @function collectAllInvalidWords and
- * @function removeDuplicates.
+ * Bundling the functionalities of @function collectAllInvalidWords and @function removeDuplicates.
```
TimothyGillespie: The @function keyword is not supposed to be used like this. Rather it serves as an indicator that an object, wie might not appear to be a function, is actually a function:
https://devdocs.io/jsdoc/tags-function
An example could be `const iLookLikeAVariable = () => { // Do Something};`

TimothyGillespie:
Seems to fix the issue as I couldn't reproduce it.

I will approve this when

- the new test cases for this error will be added (and maybe also the previous one for $Sonne Sonne$ which I've forgotten last time).
- the JSDocs with the @function keyword will be edited or reasoned why the keyword is used correctly here.
- the reason for the issue could be described very briefly and how it was fixed. (If this is easily describable in no more than three sentences)

Korosensei42:
@TimothyGillespie I think I adressed everything you asked for. For a more detailed explanation see the edited description at the top. Thanks for your feedback. If there are still some problems you see, don't hesitate to tell me.

TimothyGillespie:
Looks good so far.
I think there has been a misunderstanding the dash in the JSDocs. I think it should be there. Not that it was inconsistent. Now I am not sure if it is really important. I just always see the dash put that way and it seems standard. But since JSDocs will/could be parsed later a missing dash could break it? Just in case.

I noticed one more thing which is unrelated to this specific issue. That is in line 128 in the tests. It should check for the input a space. But currently it only empties the list and then confirms that the issue list is now empty.

Anyway: good job.

If you could take a look at the test in line 128 and either explain or correct it I will give the approval for the PR.

Korosensei42:
@TimothyGillespie That most likely was a copy-paste error from some other time. I corrected it.

Issue #73, opened 2019-07-04
implement a library of exercises


RonjaKaehne:
we should implement a library of exercise tasks because of two reasons. The first reason is we have the possibility to do some more test for our proofchecker. the second reason is the user can start using our tool to exercise even without external tasks.

PR #84, opened 2019-07-22
Ronja kaehne/example page

RonjaKaehne:


Korosensei42:
In general I really like the idea of giving an explanation only when hovering over it. 

However, it is not very intuitive that this option even exists. The user may only encounter this
functionality if he either marks the whole text or hovers over the explanation by himself, which seems random when you don't know about this functionality.

I would suggest to change this to something similar to the way Discord handles it. See for example here:
https://gyazo.com/a078f1c07440bf41d335287983266c3b

The only thing I would change about this Discord-way is that we should have the ability to hide text again, just like you are doing right now. So I would only suggest some highlight of the region that can be uncovered by hovering over it.


Other than this, this looks fine. So I would approve this PR if you comment on my remark

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+class ExamplesPropositionalLogic extends React.Component {
+	public render() {
+		return <div className={styles.site}>
+			<div className={styles.proofEditor}>
+				<p><button className={styles.buttons}
+					onClick={this.showElement}>
+					Beispiel 1
+						</button></p>
+				<table className={styles.toggleTable} id="toggleTable">
```
TimothyGillespie: Choosing an example via a button is valid. But it shouldn't be done via CSS when working in react. The I created just now doesn't contain a suggestion how to do this in react yet, but I can help you with this too. In the end it will make things a lot easier if understood.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import styles from "./propositionalLogic.module.scss";
+
+class ExamplesPropositionalLogic extends React.Component {
+	public render() {
+		return <div className={styles.site}>
+			<div className={styles.proofEditor}>
+				<p><button className={styles.buttons}
+					onClick={this.showElement}>
+					Beispiel 1
```
TimothyGillespie: This should be a more descriptive name. Possible what statement is being proven in this example.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+					<tr>
+						<th>Eingabe</th>
+						<th>Erklärung (mit der Maus drüberfahren zum anzeigen)</th>
+					</tr>
+					<tr>
+						<th>Wir zeigen: a -> (b -> c) -> (a und b) -> c.</th>
+						<th>Dies soll bewiesen werden. </th>
+					</tr>
+					<tr>
+						<td></td>
```
TimothyGillespie: It looks good, but it could potentially be solved differently. This way it is an irregularity which can make it harder to process this structur. But I guess the way it is done here is okayish in this case. Not an issue I would block an merge for.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+						<td></td>
+					</tr>
+					<tr>
+						<td>Es seien a, b und c Aussagen.</td>
```
TimothyGillespie: This can be rendered more dynamically. In PR #92 there is a dynamical rendering of these so all the tr and td won't need to be retyped and copied etc.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+						<td>qed </td>
+					</tr>
+				</table>
+
+			</div>
+
+		</div >;
+	}
+
+	private showElement = async (): Promise<void> => {
```
TimothyGillespie: This is a valid way to do this outside of react. I suggested this at first in last Thursday call, but I didn't think this through with react in mind and jumped to this intuitively. You can see the react way of doing these kind of things in #92 too.
There are certain issues associated with this usage and it decentralizes the knowledge base. If you have more questions about it or why it shouldn't be used in a react environment feel free to ask.

src/components/propositionalLogic/proofEditor.tsx
```diff
@@ -20,7 +22,7 @@ class ProofEditor extends React.Component<{}, State> {
 		return <div className={styles.proofEditor}>
 			<textarea
 				className={styles.textInput}
-				placeholder="Gebe hier deinen Beweis ein."
+				placeholder={textboxPlaceHolder}
```
TimothyGillespie: Single Source Of Truth. Good!

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -2,75 +2,72 @@ import React from "react";
...
-		];
 
-		this.state = { proofPart };
+		this.state = { exampleProofList };
 	}
 
 	public render() {
 		return <div className={styles.site}>
-			<table>
+			<button>Nächstes Beispiel</button>
```
TimothyGillespie: This would need an onClick event so it might look like this:
`<button onClick={showNextExample()}`
And the function would need to reset the "clicked" array and increase the "exampleToShow" variable by one. These changes have to be put into the `this.setState({...})` methode. Generally you only wanna call it when you change a variable in the state which should change the look of the website.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -2,75 +2,72 @@ import React from "react";
...
-					{this.renderProof()}
-					<tr>
-						<td></td>
-						<td></td>
-					</tr>
-					<tr>
-						<td>Also gilt $a -> (b -> c) -> (a und b) -> c$.</td>
-						<td>qed </td>
-					</tr>
+					{this.renderProof(0)}
```
TimothyGillespie: You can use `this.state.exampleToShow` as a parameter then or just use it in the function and get rid of the parameter.

src/components/propositionalLogic/examplePage.tsx
```diff
@@ -2,75 +2,72 @@ import React from "react";
...
 				</tbody>
 			</table>
 		</div>;
 	}
```
TimothyGillespie: And the methods below this need use the new logic then.

RonjaKaehne:
Hey,
i added a new structure that allows us to pick and show predefined examples by pressing a button. we have to add some other example and should store them into a seprated file (see #73 ) so we could acces them by there. check out the functionality of the example generator :) actually the code pick a random example from all existing examples in the list. we also can increment the list by each cick on the button for sure. this one was just a possible suggestion

Korosensei42:
I'm gonna get to #73 tomorrow, most likely

TimothyGillespie:
Issue #98 and #99 will appear with the pull request.

TimothyGillespie:
@Korosensei42 Please have a quick look over this and merge it if you seem fit. You could get started on #99 if you wish to do so then.

Korosensei42:
@TimothyGillespie I thought I might as well do #99 now and merge afterwards. This way we don't need an additional PR.

Regarding merging this: I get 3 errors in Diproche/github/webinterface/src/components/propositionalLogic/examplePage.tsx., so merging would not feel appropriate right now. Do you have a suggestion on how to fix them?

TimothyGillespie:
What errors are you getting? I don't see any and the CLI says it's passing too?

Korosensei42:
I get three errors:

1. In line 5 with exampleProofList.
Currently it looks like this:
```
const exampleProofList: Array<Array<[string, string]>> = data.examples;
```
Error description: 
```
Type '((string | string[])[] | (string | string[])[][])[][]' is not assignable to type '[string, string][][]'.
  Type '((string | string[])[] | (string | string[])[][])[]' is not assignable to type '[string, string][]'.
    Type '(string | string[])[] | (string | string[])[][]' is not assignable to type '[string, string]'.
      Type '(string | string[])[]' is missing the following properties from type '[string, string]': 0, 1ts(2322)
```

2. In line 28 with exampleNumber, false (arguments of renderProof).
Currently it looks like this:
```
	public render() {
		return <div className={styles.site}>
			{this.renderExampleButtons()}
			<table className={styles.example} id="solve">
				{this.renderProof(exampleNumber, false)}
			</table>
		</div >;
	}
```

Error description:
```
Expected 0 arguments, but got 2.ts(2554)
```

3. In line 61 with `proofPart`.

Currently it looks like this:
```
			renderedPart.push(<tr>
				<td>{line[0]}</td>
				<td className={scssClass} onClick={() => this.showExplanation(index, proofPart)}>{explanation}</td>
			</tr>);
```

Error Description:
Expected 1 arguments, but got 2.ts(2554)


Now, fixing 2 and 3 would be no big deal, I guess. However, I don't know what would be correct for 1.

I also pulled, so I don't get why only I get these errors.
Btw: Sorry for the wall of text, but I figured I'd better be as detailed as possible.

TimothyGillespie:
The errors should be fixed now. I don't know why I didn't get them myself first. I removed the lines with `Aufgabe: ...` since the initial idea would be to have this be 1:1 typable into the proofEditor and work and also the first line should state what we are trying to prove. These are not supposed to be task yet but simply full examples. The tasks will be on another page. If you think they need an header we can still add one. @Korosensei42 did you intend to use your addition into the json as predefined tasks?

Korosensei42:
@TimothyGillespie It doesn't really matter if we include them here, since, as you said, we can still create a headline if we want to use the examples as tasks. So deleting them here is totally fine with me,

PR #86, opened 2019-07-25
Ronja kaehne/fix bracket bug#85

RonjaKaehne:


src/util/input_formatter/regexes.ts
```diff
@@ -1,5 +1,7 @@
 export const Regexes = {
```
TimothyGillespie: Some general advice one RegExps:
When you make a group but you don't need to fetch the specific content of the group you can make it a non-capturing group with ?:
example
Capturing Group: `(this|or|that)`
Non-Capturing group: `(?:this|or|that)`

The main benefit is it can make it easier to handle certain functionalities of RegExps, make clear that you won't be fetching this group content later and it should also improve the runtime.

I am not sure if it could be applied to any of these here and it wouldn't be worth to revise them all. It just came to mind right now.

TimothyGillespie:
Also: If you reference the issue in the branch and PR-Name it's rather unwiedly since you cannot click on it and have to seek it out manually. It also won't show up in the issue itself. My advice would be to just point out what the PR/Branch is supposed to change(FixBracketBug works for me) and put the issue reference number in the description like this: #85 
Now this PR will show up in the issue and it's clickable.

PR #92, opened 2019-07-26
Enable display of explanation by clicking

TimothyGillespie:
This works already. It seems like I have broken the design a little bit, but I cannot find where or how immediately.
This is easier to maintain and expand.


TimothyGillespie:
Also enables maintainability and enhancability.

Issue #91, opened 2019-07-26
Fix tslint issues


TimothyGillespie:
Fix the remaining tslint issues in #77

Korosensei42:
I adressed all the tslint issues. However, semaphoreci still isn't happy. For more details see my last comment in #77

Korosensei42:
This was adressed and fixed in #77

PR #93, opened 2019-07-27
Remove unused test

### LINKED ISSUE #90: Remove the now unused test in the vocabchecker tests
As has been mentioned last Thursday there is one test which seems to have lost it's meaning and it only tests input_formatter function by now. It should be removed since it would just be a duplicated test and lead to confusion.

Korosensei42:
This should fix #90  and #91.


Korosensei42:
There don't seem to be any lintissues in this branch. At least VSCode doesnt display any such issues to me when executing `npm run lint`. So this only fixes the problem with the unused tests. For the other problem I most likely have to check out the branch mentioned in #77.

src/util/vocabularyChecker/detectWrongSyntax.test.ts
```diff
@@ -111,6 +111,7 @@ describe("logSingleWord", () => {
 
 describe("CollectAllInvalidWordsInIssues", () => {
 	it("Creates a correct Issue-array for one wrong word containing only this word", () => {
+		emptyIssueList();
```
TimothyGillespie: You can use the beforeEach function with these. If it is being used in a describe block it will only be called before each tests inside the describe block.
https://jestjs.io/docs/en/setup-teardown.html

Korosensei42: Good suggestion! I added the beforeEach now. However, the approval got stale because of that. Could you renew it again?

TimothyGillespie: Sure thing.

PR #94, opened 2019-07-30
Kweirauch/integrate exercises as json

Korosensei42:
This later aims to solve #73. Right now only a very naive and basic approach is implemented.

### DIFF
--- a/src/components/propositionalLogic/example.json
+++ b/src/components/propositionalLogic/example.json
@@ -1,6 +1,6 @@
 {
     "example1": [
-        ["Aufgabe: Zeige a -> (b -> c) -> (a und b) -> c", "", true],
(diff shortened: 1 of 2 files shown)


src/components/propositionalLogic/example.json
```diff
@@ -0,0 +1,50 @@
+{
+    "example1": [
```
TimothyGillespie: I just saw here that you already got to #99. In the other PR I just named the file examples.json with the plural form I think would be more fitting since they are multiple examples. If you remove the boolean values and bring them in the form of the examples.json from PR #84 then Issue #99 will be solved.

Korosensei42:
This is outdated now.

Issue #98, opened 2019-07-31
Have the change example buttons not stack upon each other


TimothyGillespie:
Rather they should be put next to each other.

TimothyGillespie:
Assuming localhost on port 3000 the address would be http://localhost:3000/aussagenlogisches_beweisen/beispiele

TimothyGillespie:
Solved by #100

PR #100, opened 2019-07-31
Exercises

TimothyGillespie:
This will need some work still. The proofEditor should be reused in here and will probably need some externalization. Solves #98 too.


TimothyGillespie:
This is almost ready.
During the conversion I changed some file structurs which broke some CSS Styling. This will need to be fixed.
Also I want to make exercise a controlled component as well.

With these two changes and some clean up this should be ready to go.

TimothyGillespie:
This pull request does also:
- Fix an issue for the vocab checker because it would only collapse one `\n` instead of all of them
- Make the reduced design changes as has been agreed upon

Korosensei42:
Looks good to me so far. I will try out some dropdown menu for a bit, though. If I think it looks good, I will suggest it, otherwise I will just approve this

PR #101, opened 2019-08-01
Kweirauch/fix vocab bug

### LINKED ISSUE #95: Unkown vocabulary will not create an issue when placed immediately in front of a fullstop
I couldn't get the most up to date version of our master branch as of now, so please make sure to check if does reproduce.

### Current Behavior
For the input `Die Sonne.` there will only be an issue listed for "Die", but not for "Sonne".
Generally: It doesn't seem to produce issues for unkown words right in front of a fullstop. If there is a white space inbetween the fullstop and the word it will create an issue.

### Expected Behavior
For the input `Die Sonne.` it will throw an issue for the word "Die" and "Sonne".

Korosensei42:
This should fix #95.

Previous behaviour:

Inputs like `Die Sonne.`, i.e. illegal words before interpunctuation did not create an issue.

Exprected behaviour:

Such inputs should create an Issue.

How I solved it: Before even searching for any illegal words I replace every interpunctuation with a whitespace. This is done by `text = text.replace(/[.,\/#!$%\^&\*;:{}=\-_`~()]/g, "");`, currently in line 24 in detectWrongSyntax.ts.

Additional stuff: Formulated some of the documentation differently, since it wasn't on point for me.

@TimothyGillespie I already played around with some inputs. Could you test some, too? Just in case I overlooked something. By the way: The regex above has the advantage, that now even expressions like `$Sonne$` correctly throw issues now. This means, we are restricting propositional variables to only be allowed words, too. However, the indexing does not take in account interpunctuation. That is an issue that I should solve in another PR later, though.


src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -6,7 +6,7 @@ import uniqueValues from "./iterators";
 
 /**
  * Regular expression that looks for any and all words. A word is defined as some sequence of
- * characters (\w+) betweeen wordbarriers (\b)
+ * characters (\w+) betweeen wordbarriers (\b).
```
TimothyGillespie: betw**eee**n has one e too many.

src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -16,10 +16,11 @@ const allowedWords = json;
 export interface Position { fromIndex: number; toIndex: number; }
 
 /**
- * Adds an Issue for every invalid word.
+ * Adds an Issue for every invalid word from a given text.
```
TimothyGillespie: Should be: **in** a given text.

src/util/vocabularyChecker/detectWrongSyntax.ts
```diff
@@ -87,7 +88,7 @@ function logMultipleOccurences(word: string, position: Position[]): Issue[] {
 }
 
 /**
- * Creates an Issue of one word.
+ * Creates an Issue of a single word.
```
TimothyGillespie: **for** a single word

Korosensei42:
@TimothyGillespie I adressed your feedback.

> This doesn't seem to work quiet yet. Your console.log seems to find the right start positions, as far was I can tell, but it seems no issue will be created for them.

Could you tell me on which inputs it does not work as expected? I didn't stumble upon any additional problems.

TimothyGillespie:
For any input it seems.

For example "Die Sonne."
But also "Die Sonne" without punctuation.

My working browser is firefox on a windows machine. I will test again soon if I can reproduce it.

Korosensei42:
That is weird. For me Issues get created and displayed, even though I am on Firefox as well.

TimothyGillespie:
This change fixes the issue for me.

When throwing an error a function won't give a return. This caused to give no Issue when back. So when no error list has been found in the diproche return it will now just pretend there is no error given by diproche and the function can return as usual.

I didn't get this issue beforehand either and I wonder why this hasn't been the same for you, but this should work for the both of us now.

Issue #108, opened 2019-08-12
React Router Doesn't Work When Not Served


TimothyGillespie:
Since the user should later be able to just open a html file there will be no server serving the file and using react router won't work a file path.

This could be switched over to a State.

TimothyGillespie:
A potential solution can be found in #109 but it won't be merged since other issues arise.

TimothyGillespie:
Related issue can be found in #110

TimothyGillespie:
Will be solved if swipl-wasm can be utilized since it can then just be served by a host.

PR #109, opened 2019-08-12
Convert React Routers To A State Based System

TimothyGillespie:
Solves #108.
Other issues arise though. Without serving apparently cannot be used with the prolog server and the build won't throw vocab issues either.


TimothyGillespie:
A potential future solution if the rest of the issues can be resolved caused by not serving it.

PR #112, opened 2019-08-16
Create an asynchronous wrapper around the swipl-wasm-interface

Khazuar:
An attempt to make the prolog connection easier to use.

Does not handle multiple results or errors correctly yet.

### DIFF
--- a/src/util/prolog/fubar.test.ts
+++ b/src/util/prolog/fubar.test.ts
@@ -1,14 +1,59 @@
 import "../../../swipl-wasm/swipl-web";
-import { loadingFinishedPromise, query } from "./fubar";
+import { consult, query } from "./fubar";
 
 describe("Fubar", () => {
-	test("Tests bob is a man", async () => {
-		await loadingFinishedPromise;
-		query("man(bob).");
+	test("Test whether bob and frank are men", async () => {
+		await consult("man(bob).");
+		expect(await query("man(bob).")).toEqual(["true."]);
+		expect(await query("man(frank).")).toEqual(["false."]);
 	});
 
-	test("Tests frank is not a man", async () => {
-		await loadingFinishedPromise;
-		query("man(frank).");
+	test("Waiting on previous results and maintaining order is not necessary", async () => {
+		consult("man(bob).");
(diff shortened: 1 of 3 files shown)


Khazuar:
The PR might go into another target branch instead.

TimothyGillespie:
It surely does help. 

One issue I encounter is, that the first test will take to long and time out. This problem has occured with my branch as well. I cannot see why this behavior happens though. I think this is beyond my ability as of now.

It seems that swipl-wasm cannot handle multiple results in general. It will recognize that something is yet to come but I couldn't find a way to get a second result on their example page too.
But this wouldn't be necessary for our purpose and could be resolved with the bagof/3 and findall/3 predicates if multiple answers are needed.

Khazuar:
> It surely does help.
> 
> One issue I encounter is, that the first test will take to long and time out. This problem has occured with my branch as well. I cannot see why this behavior happens though. I think this is beyond my ability as of now.

On your local machine as well or on semaphore only? Locally this works for me. I can only speculate to the reason behind the issue on semaphore:
* maybe the setup there doesn't support wasm
* maybe we did something wrong
* maybe it just takes longer there because our local machines are much faster.
Either way, this should be fixable with a little more experimenting. I tend to the last explanation, since I see locally that the first test is taking a hundred times longer than the subsequent ones - that's the initial loading of the wasm engine and of SWI. 

> It seems that swipl-wasm cannot handle multiple results in general. It will recognize that something is yet to come but I couldn't find a way to get a second result on their example page too.

Yes. I tried for multiple hours to figure out how to get the next results out. I found [this page in the SWI manuals](https://www.swi-prolog.org/pldoc/man?section=calling-prolog-from-c) that describe two ways to work with the Foreign Language Interface of SWI, either by PL_call (which we use right now and which is described in there as simple, but limited) and another one, where you need to call multiple different specialized methods. I didn't try the other one actually. 

> But this wouldn't be necessary for our purpose and could be resolved with the bagof/3 and findall/3 predicates if multiple answers are needed.

I agree, since we're not running arbitrary queries, and can just work around the issue anyways, the current solution might be just sufficient.

PR #114, opened 2019-08-17
Add $-signs to examples and exercises

### LINKED ISSUE #113: Make notation in exercises and examples uniform
Currently, for the examples there is mixed notation. I.e.
sometimes there are $-signs around formulas and sometimes there are not. 
Additionally the reader might be confused when encountering them, since it might
feel unnatural for non-LaTeX-conform people. So all examples and exercises
should not contain $-signs.

Korosensei42:
This fixes #113 by removing the $-signs from the formulas.

### DIFF
--- a/src/components/propositionalLogic/examples.json
+++ b/src/components/propositionalLogic/examples.json
@@ -1,47 +1,47 @@
 {
   "examples":[
         [
-          ["Wir zeigen: a -> (b -> c) -> (a und b) -> c.", "Dies soll bewiesen werden."],
-          ["Es seien a, b und c Aussagen.", "Allgemeine Vorraussetzung"],
-          ["Angenommen a ->(b -> c).", "Vorraussetzung: Die Prämisse a -> (b -> c) ist wahr."],
-          ["Angenommen ferner es gilt (a und b).", "Vorraussetzung: (a und b) ist wahr."],
-          ["Dann folgt a.", "a folgt, da (a und b) nur dann wahr ist, wenn a wahr ist."],
-          ["Ausserdem folgt b.", "b folgt, da (a und b) nur dann wahr ist, wenn b wahr ist."],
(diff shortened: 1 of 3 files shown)


TimothyGillespie:
See my note in #113.

The example should contain the exact way you would enter it in the sandbox. If you remove them it won't work entered 1:1.

Korosensei42:
I changed it accordingly and also clarified some notation in the wiki.

src/components/propositionalLogic/examples.json
```diff
@@ -1,47 +1,47 @@
...
-          ["Wir zeigen: a -> (b -> c) -> (a und b) -> c.", "Dies soll bewiesen werden."],
-          ["Es seien a, b und c Aussagen.", "Allgemeine Vorraussetzung"],
-          ["Angenommen $a ->(b -> c)$.", "Vorraussetzung: Die Prämisse a -> (b -> c) ist wahr."],
-          ["Angenommen ferner es gilt $(a und b)$.", "Vorraussetzung: (a und b) ist wahr."],
-          ["Dann folgt a.", "a folgt, da (a und b) nur dann wahr ist, wenn a wahr ist."],
-          ["Ausserdem folgt b.", "b folgt, da (a und b) nur dann wahr ist, wenn b wahr ist."],
-          ["Damit gilt $(b -> c)$.", "Die Konlusion (b -> c) gilt, da die Prämissen a,sowie (a und b) beide gelten."],
-          ["Ferner folgt c.", "Die Konlusion c gilt, da die Prämissen a, sowie (a und b) beide gelten."],
+          ["Wir zeigen: $a -> (b -> c) -> (a und b) -> c$.", "Dies soll bewiesen werden."],
+          ["Es seien $a, b und c$ Aussagen.", "Allgemeine Vorraussetzung"],
```
TimothyGillespie: Here is still a $ around the enumeration. I think in an enumaration it should be seperate since otherwise the "und" will be logical?

Korosensei42: Yes, the "und" here has other semantics. Good job!

PR #120, opened 2019-08-22
Implement success message

TimothyGillespie:


TimothyGillespie:
Semaphore has no issue with this PR now while it displayed some more additional npm install error on the wasm branch. Maybe this could be some indication for something.

TimothyGillespie:
This will fix #120

TimothyGillespie:
This will fix #120

TimothyGillespie:
See the comments in #121. It will render the success message as soon as something is being typed in and not wait for the button push.

TimothyGillespie:
You're right about that. I now did solution 2 with an "empty" issue. This seems to work better. It will render it only after the proof has been checked now.

